PR #16, opened 2017-12-05
add SjpegFreeBuffer(…) to the library API so that it could be used not only in C++

magicgoose:
Languages other than C++ do not have `delete[]` equivalent.
I think it would be a lot easier to use the library correctly with this patch.


skal65535:
Thanks for the suggestion! 
I've committed a similar version as https://github.com/webmproject/sjpeg/commit/957445eff2145e9a7a91676cc1d0c09656d61e9c

PR #99, opened 2020-02-20
Print bpp along with number of bytes

skal65535:


skal65535:
should be ok. That message is just informative.

PR #103, opened 2020-12-12
Use ${LIB_INSTALL_DIR} to specify library installation dir

eclipseo:
Most system use lib64 to install libraries on 64 bits arch system.


skal65535 (approved):
Thanks for the fix!

skal65535:
Note: https://github.com/webmproject/sjpeg/pull/104 will go further and make use of GNUInstallDirs variables

PR #110, opened 2021-04-22
CI: Create AppVeyor configuration

EwoutH:
Creates an AppVeyor configuration to build sjpeg with CMake and Visual Studio 2019 (MSVC) for the x64 and ARM64 platforms.

An example build can be [found here](https://ci.appveyor.com/project/EwoutH/sjpeg/builds/38824899). AppVeyor needs to be enabled on the [GitHub Marketplace](https://github.com/marketplace/appveyor).

Currently the Arm64 job fails. See #109.

PR #113, opened 2021-04-22
Add basic CI configuration for Linux, macOS and Android

EwoutH:
Adds a basic CI configuration using GitHub Actions which builds sjpeg:
 - on Linux using CMake and GCC 10
 - on macOS using CMake and AppleClang 12
 - for Android using CMake and the Android NDK

This configuration uses GitHub Actions, which needs to be [activated here](https://github.com/webmproject/sjpeg/actions). An example build from my fork can be [seen here](https://github.com/EwoutH/sjpeg/actions/runs/775121110).

For Windows there is already an AppVeyor configuration, which was added in #110.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,30 @@
+name: Build
+on: [push, pull_request]
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    env:
+      CC: gcc-10
+      CXX: g++-10
+    steps:
+      - uses: actions/checkout@v2
+      - run: cmake .
+      - run: cmake --build .
+
+  macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: cmake .
+      - run: cmake --build .
+
+  android:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: mkdir build && cd build
+      - run: cmake .. -DCMAKE_TOOLCHAIN_FILE=../cmake/android.cmake -DSJPEG_ANDROID_NDK_PATH=$ANDROID_NDK_LATEST_HOME
+        working-directory: build
+      - run: cmake --build .
+        working-directory: build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,21 +8,21 @@ jobs:
       CC: gcc-10
       CXX: g++-10
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


EwoutH:
@jzern @madebr @skal65535 I'm doing some cleanup of old, open PRs. If you would like to get these changes in, please leave a reply or review.

In two weeks I will close this PR if no reaction received.

jzern:
Thanks for pinging this, I guess it got missed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+      - run: cmake --build .
+
+  macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: cmake .
+      - run: cmake --build .
+
+  android:
```
jzern, line 22: Is there a way to run this before merging? Elsewhere I've seen the [NDK manually installed](https://github.com/AOMediaCodec/libavif/blob/main/.github/workflows/ci-android-jni.yml#L12), is there one available by default?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: mkdir build && cd build
```
jzern, line 26: it's fine either way, but any reason this is done in a separate directory, but not for the others?

.github/workflows/ci.yml
EwoutH: Yes, you can either [activate GitHub Actions here](https://github.com/webmproject/sjpeg/actions) or see [this example run](https://github.com/EwoutH/sjpeg/actions/runs/3341346795) on my fork.

As for the Android NDK, it's using the default one installed on the GitHub Actions [runner-images](https://github.com/actions/runner-images). Currently that's the [NDK 25.1.8937393](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#environment-variables-3).

.github/workflows/ci.yml
EwoutH: Good question. Probably because of some error I encountered while testing this jobs. I don't remember exactly.

I don't see anything weird in the build logs, so I don't think it matters much.

.github/workflows/ci.yml
jzern: Thanks for the pointers. @vigneshvg maybe the extra NDK install step isn't necessary in libavif.

jzern (approved):

PR #115, opened 2021-05-19
Move cmake files from datadir to libdir

eclipseo:
Current location is preventing me to detect it correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,7 +209,7 @@ write_basic_package_version_file(
 
 # Create the Config file.
 include(CMakePackageConfigHelpers)
-set(ConfigPackageLocation ${CMAKE_INSTALL_DATADIR}/sjpeg/cmake/)
+set(ConfigPackageLocation ${CMAKE_INSTALL_LIBDIR}/cmake/sjpeg/)
 configure_package_config_file(
   ${CMAKE_CURRENT_SOURCE_DIR}/cmake/sjpegConfig.cmake.in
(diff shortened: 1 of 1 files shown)


jzern (commented):
I think either path will work [1] and there's nothing architecture specific in the file. Did you run into an issue with the existing path?

[1] https://cmake.org/cmake/help/v3.19/command/find_package.html#search-procedure

eclipseo:
It might be specific to my build setup: I use RPM to build a package and it detects the cmake folder automatically in lib64/cmake and thus adds an automatic Provides: cmake(nameofthelibrary) to be called from other packages depending on it. But it does not work in /usr/share. This is pretty distribution specific so I can keep my patch downstream and close this PR if it's ok.

jzern:
Thanks for the background. It looks like there isn't a canonical location for these though the cmake docs use lib/ in their current examples [1]. Out of curiosity do any of the other forms work like changing sjpeg/cmake to cmake/sjpeg under /usr/share?

[1] https://cmake.org/cmake/help/v3.20/manual/cmake-packages.7.html?highlight=configpackagelocation#creating-packages

eclipseo:
I've tested cmake/sjpeg under /usr/share and it works. I don't know the list or all working pattern, nor how they are detected by RPM.

jzern:
Thanks for trying it out. I'll let Pascal comment, but staying with /usr/share is fine to me.

skal65535:
thanks for trying both solutions. If they are both working and allowed, let's just keep it as is, to minimize disruption!

skal65535 (commented):
i think we should just delete this obsolete PR

PR #117, opened 2022-10-11
cmake: fix target name collision with other cpufeatures library

madebr:
When using a vendored sjpeg, the `cpufeatures` target clashes with other `cpufeatures` targets.
Avoid this by adding a `-sjpeg` suffix.

This happens when building SDL_image with vendored libraries.
SDL_image has support for libjxl, which uses sjpeg as a 3rd party library.
(libwebp also provides a cpufeatures target)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,11 +46,11 @@ set(PROJECT_VERSION 0.1)
 
 if(ANDROID)
   include_directories(${SJPEG_ANDROID_NDK_PATH}/sources/android/cpufeatures)
-  add_library(cpufeatures
+  add_library(cpufeatures-sjpeg
     STATIC ${SJPEG_ANDROID_NDK_PATH}/sources/android/cpufeatures/cpu-features.c
   )
-  target_link_libraries(cpufeatures dl)
-  set(SJPEG_DEP_LIBRARIES ${SJPEG_DEP_LIBRARIES} cpufeatures)
+  target_link_libraries(cpufeatures-sjpeg dl)
+  set(SJPEG_DEP_LIBRARIES ${SJPEG_DEP_LIBRARIES} cpufeatures-sjpeg)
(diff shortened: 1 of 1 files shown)


skal65535 (approved):

jzern (approved):

jzern:
Out of curiosity, how are the targets imported in this case, via include_directories()?

madebr:
In [SDL_image](https://github.com/libsdl-org/SDL_image/), we add the vendored libraries using `add_subdirectory`.
[Here](https://github.com/libsdl-org/SDL_image/blob/434134702e31431c19c0a5e6de8f8383faf6dc5b/CMakeLists.txt#L436), libjxl is included
No [`ExternalProject`]()https://cmake.org/cmake/help/latest/module/ExternalProject.html) is used. Perhaps we should?

jzern:
> In [SDL_image](https://github.com/libsdl-org/SDL_image/), we add the vendored libraries using `add_subdirectory`.

Thanks, I realized after posting I meant that one rather than include_directories. I have the same change for [libwebp](https://chromium-review.googlesource.com/c/webm/libwebp/+/3946724).

> [Here](https://github.com/libsdl-org/SDL_image/blob/434134702e31431c19c0a5e6de8f8383faf6dc5b/CMakeLists.txt#L436), libjxl is included
> No [`ExternalProject`](https://cmake.org/cmake/help/latest/module/ExternalProject.html) is used. Perhaps we should?

I haven't used this one; I've seen [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html). For installed libraries `find_library` can work and give namespaced targets, but I don't think that applies here.

madebr:
Thanks for taking care of webp. I applied [a very similar patch](https://github.com/libsdl-org/libwebp/commit/12cf7b315fb7a9df56fa6b8ee30e36312659d63c) to our slightly patched fork.

Yeah, our build systems supports both system libraries and vendored libraries.
It's there for those old-school Windows developers refusing to use a package manager.
Though, it's useful now while I am creating Android prefab packages for the SDL libraries.

PR #118, opened 2023-05-06
Updates utils target to sjpeg-utils to avoid clashing with user's libs

ricardobtez:
This allows the users of the library to also compile the library and create target called utils.


jzern:
Thanks for the patch. This looks OK and should be able to land after the CLA is signed.

ricardobtez:
Hello @jzern . It appears it only needs a code owner to approve the PR. Could you help me with that?

jzern:
Thanks for signing the CLA. The checks appear to pass.

PR #119, opened 2023-06-06
Fix search-to-target-size mode.

szabadka:
If the binary search did not find a best result in the last step, then the stored run levels are invalid and have to be regenerated.

### DIFF
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -123,6 +123,7 @@ void Encoder::LoopScan() {
   float best = 0.;     // best distance
   float best_q = 0.;  // informative value to return to the user
   float best_result = 0.;
+  bool last_is_best = false;
   for (int p = 0; p < passes_; ++p) {
     search_hook_->pass = p;
     // set new matrices to evaluate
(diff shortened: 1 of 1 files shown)


src/dichotomy.cc
```diff
@@ -167,18 +167,18 @@ void Encoder::LoopScan() {
   SetQuantMatrices(opt_quants);
   for (int c = 0; c < 2; ++c) FinalizeQuantMatrix(&quants_[c], q_bias_);
 
-  // return informative values to the user
-  search_hook_->q = best_q;
-  search_hook_->value = best_result;
-
   // optimize Huffman table now, if we haven't already during the search
-  if (!search_hook_->for_size) {
+  if (!search_hook_->for_size || best_q != search_hook_->q) {
```
skal65535, line 171: at line 126, we could declare:
  bool last_is_best = false;

and at line 154, we could have:

last_is_best = (p == 0 || fabs(result - search_hook_->target) < best);
if (last_is_best) {
 ...
}

Then this line would just be: if (!search_hook_->for_size || !last_is_best) {
   ...
}


WDYT?

szabadka: Yes, this is more explicit, thanks. Updated the PR.

skal65535 (approved):
Thanks for he fix!

PR #122, opened 2023-09-22
CMakeLists: bump minimum cmake version to 2.8.12

jzern:
quiets a warning with newer CMake versions:

Compatibility with CMake < 2.8.12 will be removed from a future version
of CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 2.8.12)
 
 # Prefer GLVND libraries by default.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 2.8.12)
```
jzern, line 15: I didn't check on the compatibility of the second half of the message:

```
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

skal65535 (approved):

PR #127, opened 2024-05-20
EncoderNV12: remove unused W_/H_ private members

jzern:
This matches the other classes that derive from Encoder.


jzern:
Maybe there was some patch desync.

PR #130, opened 2024-11-06
add missing <cstdlib> include

jzern:
needed for uses of `std::abs`

PiperOrigin-RevId: 693699945

### DIFF
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -22,6 +22,7 @@
 #include <float.h>    // for FLT_MAX
 #include <stdint.h>
 
+#include <cstdlib>
 #include <mutex>  // NOLINT
 #include <new>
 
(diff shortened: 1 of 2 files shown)


src/enc.cc
```diff
@@ -22,6 +22,7 @@
 #include <float.h>    // for FLT_MAX
 #include <stdint.h>
 
+#include <cstdlib>
```
jzern, line 25: Probably some more can be done here with iwyu. I'm leaving stdlib.h in this commit to match the change, though I think it's been widely discussed that it would be nearly impossible to remove putting things in the global namespace in the C++ versions of the headers.

jzern:
cc: @vrabaud

vrabaud (approved):

PR #139, opened 2025-04-29
enc.cc: extract member vars from loops

jzern:
This focuses on member variables used in branches within loops which are
constant through the loop's execution. This avoids a reload of the value
and may allow the compiler to extract the condition and split the loops.

### DIFF
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -1182,10 +1182,11 @@ void Encoder::CollectHistograms() {
   int16_t* in = in_blocks_;
   const int mb_x_max = W_ / block_w_;
   const int mb_y_max = H_ / block_h_;
+  const bool use_extra_memory = use_extra_memory_;
   for (int mb_y = 0; mb_y < mb_h_; ++mb_y) {
     const bool yclip = (mb_y == mb_y_max);
     for (int mb_x = 0; mb_x < mb_w_; ++mb_x) {
-      if (!use_extra_memory_) {
+      if (!use_extra_memory) {
         in = in_blocks_;
       }
       GetSamples(mb_x, mb_y, yclip | (mb_x == mb_x_max), in);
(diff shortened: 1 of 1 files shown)


src/enc.cc
```diff
@@ -1231,11 +1232,12 @@ void Encoder::SinglePassScan() {
...
+  const bool have_coeffs = have_coeffs_;
   for (int mb_y = 0; mb_y < mb_h_; ++mb_y) {
     const bool yclip = (mb_y == mb_y_max);
     for (int mb_x = 0; mb_x < mb_w_; ++mb_x) {
       if (!CheckBuffers()) return;
-      if (!have_coeffs_) {
+      if (!have_coeffs) {
         in = in_blocks_;
         GetSamples(mb_x, mb_y, yclip | (mb_x == mb_x_max), in);
         fDCT_(in, mcu_blocks_);
```
jzern, line 1243: x86 can call using a pointer argument, so no gain there. Arm `blr`s from a register, but given the register pressure it would just add to the stack usage.

jzern:
Some micro optimizations while looking around in response to something else.

src/enc.cc
```diff
@@ -1527,10 +1529,12 @@ void Encoder::SinglePassScanOptimized() {
   int16_t* in = in_blocks_;
   const int mb_x_max = W_ / block_w_;
   const int mb_y_max = H_ / block_h_;
+  const bool have_coeffs = have_coeffs_;
+  const bool reuse_run_levels = reuse_run_levels_;
   for (int mb_y = 0; mb_y < mb_h_; ++mb_y) {
     const bool yclip = (mb_y == mb_y_max);
     for (int mb_x = 0; mb_x < mb_w_; ++mb_x) {
```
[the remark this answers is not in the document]
skal65535: agreed. bool flags can probably be packed , but not `mb_w_`

src/enc.cc
skal65535: agreed

skal65535:
LGTM

skal65535 (approved):

PR #140, opened 2025-04-29
enc.cc: mark some Encoder classes as final

jzern:
These classes are mostly end state and not derived from further.

### DIFF
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -1762,7 +1762,7 @@ bool FinishEncoding(Encoder* const enc, const EncoderParam& param) {
 ////////////////////////////////////////////////////////////////////////////////
 // sub-class for YUV 4:2:0 version
 
-class Encoder420 : public Encoder {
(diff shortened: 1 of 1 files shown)


jzern:
`MemorySink` is also a candidate. Just some speculative optimization.

skal65535 (approved):
yes these are totally `final`. For `MemorySink`, need to check if it's exposed to the user somehow through the API

jzern:
> For MemorySink, need to check if it's exposed to the user somehow through the API

[`ByteSink`](https://github.com/webmproject/sjpeg/blob/426d7922adb91f3a5a04a9f3266bc498abed95b0/src/sjpeg.h#L376) is, but [`MemorySink`](https://github.com/webmproject/sjpeg/blob/426d7922adb91f3a5a04a9f3266bc498abed95b0/src/bit_writer.h#L36) is in bit_writer.h. I'll take a look at that separately, bit writing is quite costly with larger resolution images.

PR #141, opened 2025-05-25
install manpages only when executables are installed.

leleliu008:


skal65535:
Closing in favor of #160, which includes this fix plus a broader CMakeLists.txt cleanup pass. Thanks for the original patch!

Issue #143, opened 2025-08-19
Modified  CMakeLists.txt and libsjpeg.pc.in


sidrat:
Modified :  libsjpeg.pc.in
------------------------------------

Please Note : pkgconfig generator omits Version number so i added it  manually and add -lsjpeg-utils to LIBS: 

cat > src/libsjpeg.pc.in<< "EOF"
prefix=@prefix@                                                                                                                                       
exec_prefix=@exec_prefix@                                                                                                                             
libdir=@libdir@                                                                                                                                       
includedir=@includedir@                                                                                                                               
                                                                                                                                                       
Name: libsjpeg                                                                                                                                        
Description: Library compressing or recompressing pictures in JPEG format.                                                                            
Version: 0.1.0                                                                                                                                        
Cflags: -I${includedir}                                                                                                                               
Libs: -L${libdir} -lsjpeg -lsjpeg-utils                                                                                                               
Libs.private: -lm @PTHREAD_CFLAGS@ @PTHREAD_LIBS@
EOF

Modified :  CMakeLists.txt
--------------------------------------

add to line 16   option(BUILD_SHARED_LIBS "Build using shared libraries" ON)

add to line  233 # Configure and install the pkg-config file                                                                                                          
add to line  234 set(prefix ${CMAKE_INSTALL_PREFIX})                                                                                                                  
add to line  235 set(exec_prefix ${prefix})                                                                                                                           
add to line  236 set(libdir ${exec_prefix}/${CMAKE_INSTALL_LIBDIR})                                                                                                   
add to line  237 set(includedir ${prefix}/${CMAKE_INSTALL_INCLUDEDIR})                                                                                                
add to line  238 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/libsjpeg.pc.in ${CMAKE_CURRENT_BINARY_DIR}/libsjpeg.pc @ONLY)                                         
add to line  239 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/libsjpeg.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig) 

This builds and install correctly on my BLFS system (Beyond Linux From Scratch). Version r12.3-1263 
sorry no diff file yet.

jzern:
Thanks for the report. There may be some useful things to discuss here, posting a PR would simplify the discussion.

> add to line 16 option(BUILD_SHARED_LIBS "Build using shared libraries" ON)

[...]

PR #146, opened 2026-03-09
build: generate sjpeg.pc pkg-config file for sjpeg (#143)

hkunz:


skal65535:
Closing in favor of #160, which adds pkg-config generation with the Requires/Libs and install-path issues fixed (libsjpeg itself doesn't depend on libjpeg/libpng — only the example tools do — and paths now go through GNUInstallDirs).

PR #147, opened 2026-07-31
Speed up SIMD JPEG quantization with a branch-free run-length scan

hasinoff:
The SSE2 and NEON quantizers detect and emit non-zero coefficients by scanning all 64 entries in zig-zag order with a data-dependent `if (v > 0)` branch. For typical (sparse) blocks this branch is highly unpredictable, so the scan stalls the pipeline and accounts for a meaningful fraction of encode time even though most coefficients are zero.

Record a natural-order non-zero bitmask while quantizing (SSE2 via `_mm_movemask_epi8`, NEON via a bit-weight reduction), remap it to zig-zag order, and iterate only the set bits using count-trailing-zeros. This removes the per-coefficient branch and touches only the handful of non-zero coefficients. Each SIMD quantizer keeps its own emission loop, so the SSE2 and NEON paths stay self-contained. The emitted run/level stream is bit-identical to the scalar QuantizeBlock() reference.

Harden the intrinsic gating while here: add a TrailingZeros64() helper that selects __builtin_ctzll via __has_builtin (rather than a bare __GNUC__ version check), with a portable fallback. CalcLog2()'s builtin gate is generalized the same way so non-GNU compilers that provide the builtin still get it.

Whole-encode benchmark (1280x960, quality 90, 4:2:0, adaptive quantization
+ Huffman optimization) improves from ~4.97 ms to ~4.17 ms per encode, about 16% faster (best-of-run; ~12% on the mean). The isolated run-length kernel is roughly 35-50% faster depending on block sparsity.

Validated on ~2,600 photographic JPEGs (re-decoded to RGB, then re-encoded at quality 90, 4:2:0, adaptive quantization + Huffman): the SSE2 output is byte-identical to both the prior SSE2 output and the scalar reference on every image (zero differences). The 32-bit ARM NEON reduction path was additionally verified bit-identical to the aarch64 path and to a scalar reference over an exhaustive set of lane masks, run natively under an armv7 emulator.

### DIFF
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -75,6 +75,20 @@ const uint8_t kZigzag[64] = {
   53, 60, 61, 54, 47, 55, 62, 63,
 };
 
+// Inverse of kZigzag: maps a natural coefficient index to its zig-zag scan
+// position (kInvZigzag[kZigzag[i]] == i). Used by the SIMD run-length emitters
+// to turn a natural-order non-zero bitmask into a zig-zag-ordered one.
+static const uint8_t kInvZigzag[64] = {
+  0,   1,  5,  6, 14, 15, 27, 28,
+  2,   4,  7, 13, 16, 26, 29, 42,
+  3,   8, 12, 17, 25, 30, 41, 43,
+  9,  11, 18, 24, 31, 40, 44, 53,
+  10, 19, 23, 32, 39, 45, 52, 54,
+  20, 22, 33, 38, 46, 51, 55, 60,
+  21, 34, 37, 47, 50, 56, 59, 61,
+  35, 36, 48, 49, 57, 58, 62, 63,
+};
+
 const uint8_t kDefaultMatrices[2][64] = {
   // these are the default luma/chroma matrices (JPEG spec section K.1)
   { 16,  11,  10,  16,  24,  40,  51,  61,
@@ -465,9 +479,16 @@ void Encoder::InitCodes(bool only_ac) {
 ////////////////////////////////////////////////////////////////////////////////
 // Quantize coefficients and pseudo-code coefficients
 
+#if defined(__has_builtin)
+#define SJPEG_HAS_BUILTIN(x) __has_builtin(x)
+#else
+#define SJPEG_HAS_BUILTIN(x) 0
+#endif
+
 static int CalcLog2(int v) {
-#if defined(__GNUC__) && \
-    ((__GNUC__ == 3 && __GNUC_MINOR__ >= 4) || __GNUC__ >= 4)
+#if SJPEG_HAS_BUILTIN(__builtin_clz) || \
+    (defined(__GNUC__) && \
+     ((__GNUC__ == 3 && __GNUC_MINOR__ >= 4) || __GNUC__ >= 4))
   return 32 - __builtin_clz(v);
 #else
   const int kLog2[16] = {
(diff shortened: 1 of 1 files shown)


skal65535:
nice work Sam! :)

skal65535:
did you try with extreme quality setting too (q=0 and q=100 e.g.), to exercise various sparseness condition? 
In particular, q=100 is interesting to see if there's an penalty cost compared to the original full-scan loop.

hasinoff:
Hi Pascal!

I redid the benchmarking on x86 with a sweep that goes all the way from q=0 to q=100 - the speedup is meaningful for all cases (at least ~22% faster), slowest for q=100 (least sparsity), and faster as quality goes down to 0 (most sparsity). Subsampling 4:4:4 instead of 4:2:0 also leads to a bit more speedup, as expected, due to there being more sparsity still.

Note that the speedups reported here are better than the PR description's claims ("about 16% faster"). That's because the numbers here are for a corpus of 2663 photographic JPEGs from Open Images, whereas the measurement for the PR description used a synthetic image (higher frequency, less realistic). 

## Subsampling 4:2:0

| Quality | OLD MP/s | NEW MP/s | Time reduction | OLD med | NEW med | Median time reduction | Byte-parity |
|--:|--:|--:|--:|--:|--:|--:|:--|
| 0 | 277.4 | 393.6 | 29.5% | 273.8 | 386.9 | 29.2% | OK (2663) |
| 10 | 249.8 | 346.4 | 27.9% | 241.8 | 333.2 | 27.4% | OK (2663) |
| 25 | 222.1 | 301.2 | 26.3% | 211.4 | 283.8 | 25.5% | OK (2663) |
| 50 | 199.2 | 266.1 | 25.1% | 189.8 | 252.0 | 24.7% | OK (2663) |
| 75 | 168.9 | 223.5 | 24.4% | 156.7 | 206.5 | 24.1% | OK (2663) |
| 90 | 139.2 | 183.1 | 24.0% | 133.5 | 175.3 | 23.8% | OK (2663) |
| 95 | 131.1 | 172.0 | 23.8% | 127.6 | 166.2 | 23.2% | OK (2663) |
| 100 | 101.5 | 131.5 | 22.8% | 101.8 | 131.8 | 22.8% | OK (2663) |

## Subsampling 4:4:4

| Quality | OLD MP/s | NEW MP/s | Time reduction | OLD med | NEW med | Median time reduction | Byte-parity |
|--:|--:|--:|--:|--:|--:|--:|:--|
| 0 | 153.0 | 222.9 | 31.4% | 152.9 | 222.5 | 31.3% | OK (2663) |
| 10 | 144.1 | 207.7 | 30.6% | 141.4 | 202.5 | 30.2% | OK (2663) |
| 25 | 131.3 | 186.0 | 29.4% | 126.9 | 178.5 | 28.9% | OK (2663) |
| 50 | 119.4 | 166.4 | 28.2% | 114.6 | 158.3 | 27.6% | OK (2663) |
| 75 | 104.0 | 142.7 | 27.1% | 98.3 | 134.1 | 26.7% | OK (2663) |
| 90 | 88.3 | 119.1 | 25.9% | 85.6 | 114.5 | 25.2% | OK (2663) |
| 95 | 83.4 | 111.5 | 25.2% | 81.5 | 108.1 | 24.6% | OK (2663) |
| 100 | 67.5 | 88.8 | 24.0% | 67.6 | 88.4 | 23.5% | OK (2663) |

skal65535:
Excellent! Impressive results, thanks for following through.
LGTM

skal65535 (approved):

PR #149, opened 2026-08-10
Add native BGRA and RGBA input paths (zero-copy, bit-exact)

hasinoff:
- Adds kBGRAInput and kRGBAInput to PixelFormat in sjpegi.h.
- Implements optimized SSE2, NEON, and C scalar block converters in colors_rgb.cc for direct packed 32-bit pixel (BGRA and RGBA) to YUV conversion, dropping alpha on the fly without intermediate RGB copies.
- Exposes sjpeg::EncodeBGRA() and sjpeg::EncodeRGBA() overloads (ByteSink* and std::string*) in sjpeg.h and enc.cc under ad-hoc specialized functions.

### DIFF
--- a/src/colors_rgb.cc
+++ b/src/colors_rgb.cc
@@ -22,6 +22,7 @@
 
 #include <string.h>
 
+#include "sjpeg.h"
 #define SJPEG_NEED_ASM_HEADERS
 #include "sjpegi.h"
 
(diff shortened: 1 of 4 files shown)


src/sjpeg.h
skal65535: this patch adds new public functions to the API surface, so is worth a minor version bump (-> 0.1.1)

src/colors_rgb.cc
```diff
@@ -22,6 +22,7 @@
 
 #include <string.h>
 
+#include "sjpeg.h"
```
skal65535: strange that you need this include, since it's also included from sjpegi.h  (line 27 below)

src/sjpeg.h
hasinoff: Done

src/colors_rgb.cc
hasinoff: Removed it - this is just the Google style guide about includes for all direct uses creeping in.

jzern: > Removed it - this is just the Google style guide about includes for all direct uses creeping in.

We can do `// IWYU pragma: export` for that.

hasinoff: Done

skal65535 (approved):
nice!

PR #150, opened 2026-08-16
Misc fixes and improvements

skal65535:
### Fixes

- **EncodeYUV420**: the clipped path read the U plane with `v_step_` — wrong output whenever the U and V strides differ.
- **dichotomy**: `StoreRunLevels()` ignored `CheckBuffers()` and wrote past `all_run_levels_` once an allocation failed.
- **bit_writer**: `Reserve()` left `buf_` dangling into memory the sink had just released.
- **SjpegEncode**: `EncoderFactory()` returns null for an unknown `yuv_mode` or on allocation failure, and the result was dereferenced unchecked.
- **Encoder**: `param.memory` was installed after construction, so blocks allocated by sub-class constructors were released by the wrong manager.
- **EncodeNV12/NV21**: arguments were validated after the base class had already used `output`, and the `uv_stride` bound was too small by half.
- **Encoder/SOF**: dimensions above 65535 were silently truncated into the SOF marker — 70000x2 encoded "successfully" as 4464x2.
- **SjpegDimensions**: rejected the null width/height that the header documents as optional, and scanned chunks with pointers built past the end of the buffer.
- **SjpegRiskiness**: the gray fraction was divided by `width * height` although the loop visits `(width-1) * (height-1)` samples, and flat tinted colors fell into the gray bucket.
- **HeaderSize**: charged two quantization matrices where `WriteDQT()` writes only one.
- **yuv_convert**: two `static bool done` guards, torn when two threads encode at once.

### Quality

- **tests**: self-contained unit-test binary, no external dependency — `make unit_test`, or `ctest` with CMake; the linux and macos CI jobs were building the tests without running them, and now run them.
- **ci**: pin `actions/checkout` to a commit hash instead of a movable tag (same version).
- **misc**: clamp the compression method instead of asserting on it, add a missing include, fix two wrong comments.

Each fix comes with the unit test that catches it.

### Note for review

Two commits may change the bitstream at byte level: **HeaderSize** (only
`YUV_400 + TARGET_SIZE`) and **SjpegRiskiness** (`YUV_AUTO`'s choice on gray
and flat tinted pictures, which is the point of the fix).

Please rebase-merge rather than squash, so the fixes stay bisectable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,7 +203,10 @@ endif()
 if(SJPEG_BUILD_TESTS)
   enable_testing()
   add_executable(unit_test ${CMAKE_CURRENT_SOURCE_DIR}/tests/unit_test.cc)
-  target_link_libraries(unit_test sjpeg)
+  # the tests encode from several threads at once
+  set(THREADS_PREFER_PTHREAD_FLAG ON)
+  find_package(Threads REQUIRED)
+  target_link_libraries(unit_test sjpeg Threads::Threads)
   add_test(NAME unit_test COMMAND unit_test)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ set(CMAKE_CXX_STANDARD 11)
 # Options for coder / decoder executables.
 option(SJPEG_ENABLE_SIMD "Enable any SIMD optimization." ON)
 option(SJPEG_BUILD_EXAMPLES "Build the sjpeg / vjpeg command line tools." ON)
+option(SJPEG_BUILD_TESTS "Build the unit tests." ON)
 
 set(SJPEG_DEP_LIBRARIES)
 set(SJPEG_DEP_INCLUDE_DIRS)
@@ -196,6 +197,14 @@ if(SJPEG_BUILD_EXAMPLES)
   target_link_libraries(vjpeg ${SJPEG_DEP_GL_LIBRARIES} sjpeg sjpeg-utils)
 
   install(TARGETS sjpeg-bin vjpeg RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+endif()
+
+# Build the unit tests if asked for.
+if(SJPEG_BUILD_TESTS)
+  enable_testing()
+  add_executable(unit_test ${CMAKE_CURRENT_SOURCE_DIR}/tests/unit_test.cc)
+  target_link_libraries(unit_test sjpeg)
+  add_test(NAME unit_test COMMAND unit_test)
 endif()
 
 # Install the different headers and libraries.
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -215,7 +215,7 @@ size_t Encoder::HeaderSize() const {
       size += (xmp_.size() / 65458 + 1) * 40;
     }
   }
-  size += 2 * 65 + 2 + 2;         // DQT
+  size += (nb_comps_ == 1 ? 1 : 2) * 65 + 2 + 2;  // DQT
   size += 8 + 3 * nb_comps_ + 2;  // SOF
   size += 6 + 2 * nb_comps_ + 2;  // SOS
   size += 2;                      // EOI
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -88,7 +88,7 @@ void Encoder::StoreRunLevels(DCTCoeffs* coeffs) {
   nb_run_levels_ = 0;
   int16_t* in = in_blocks_;
   for (int n = 0; n < mb_w_ * mb_h_; ++n) {
-    CheckBuffers();
+    if (!CheckBuffers()) return;
     for (int c = 0; c < nb_comps_; ++c) {
       for (int i = 0; i < nb_blocks_[c]; ++i) {
         RunLevel* const run_levels = all_run_levels_ + nb_run_levels_;
@@ -139,6 +139,7 @@ void Encoder::LoopScan() {
     if (search_hook_->for_size) {
       // compute pass to store coeffs / runs / dc_code_
       StoreRunLevels(base_coeffs);
+      if (!ok_) break;
       if (optimize_size_) {
         StoreOptimalHuffmanTables(nb_mbs, base_coeffs);
         if (use_trellis_) InitCodes(true);
@@ -176,18 +177,19 @@ void Encoder::LoopScan() {
   // optimize Huffman table now, if we haven't already during the search
   if (!search_hook_->for_size || !last_is_best) {
     StoreRunLevels(base_coeffs);
-    if (optimize_size_) {
+    if (ok_ && optimize_size_) {
       StoreOptimalHuffmanTables(nb_mbs, base_coeffs);
     }
   }
 
   // finish bitstream
-  WriteDQT();
-  WriteSOF();
-  WriteDHT();
-  WriteSOS();
-  FinalPassScan(nb_mbs, base_coeffs);
-
+  if (ok_) {
+    WriteDQT();
+    WriteSOF();
+    WriteDHT();
+    WriteSOS();
+    FinalPassScan(nb_mbs, base_coeffs);
+  }
   Free(base_coeffs);
 }
 
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -16,6 +16,7 @@
 //
 // Author: Skal (pascal.massimino@gmail.com)
 
+#include <assert.h>
 #include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -166,29 +167,33 @@ void Encoder::LoopScan() {
     }
     if (search_hook_->Update(result)) break;
   }
-  // transfer back the final matrices
-  SetQuantMatrices(opt_quants);
-  for (int c = 0; c < 2; ++c) FinalizeQuantMatrix(&quants_[c], q_bias_);
-
-  // return informative values to the user
-  search_hook_->q = best_q;
-  search_hook_->value = best_result;
-
-  // optimize Huffman table now, if we haven't already during the search
-  if (!search_hook_->for_size || !last_is_best) {
-    StoreRunLevels(base_coeffs);
-    if (ok_ && optimize_size_) {
-      StoreOptimalHuffmanTables(nb_mbs, base_coeffs);
-    }
-  }
-
-  // finish bitstream
+  // If the search was interrupted by a failed allocation, opt_quants[] was
+  // never filled in: there is nothing to transfer back, and nothing to write.
   if (ok_) {
-    WriteDQT();
-    WriteSOF();
-    WriteDHT();
-    WriteSOS();
-    FinalPassScan(nb_mbs, base_coeffs);
+    // transfer back the final matrices
+    SetQuantMatrices(opt_quants);
+    for (int c = 0; c < 2; ++c) FinalizeQuantMatrix(&quants_[c], q_bias_);
+
+    // return informative values to the user
+    search_hook_->q = best_q;
+    search_hook_->value = best_result;
+
+    // optimize Huffman table now, if we haven't already during the search
+    if (!search_hook_->for_size || !last_is_best) {
+      StoreRunLevels(base_coeffs);
+      if (ok_ && optimize_size_) {
+        StoreOptimalHuffmanTables(nb_mbs, base_coeffs);
+      }
+    }
+
+    // finish bitstream
+    if (ok_) {
+      WriteDQT();
+      WriteSOF();
+      WriteDHT();
+      WriteSOS();
+      FinalPassScan(nb_mbs, base_coeffs);
+    }
   }
   Free(base_coeffs);
 }
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -2154,7 +2154,7 @@ class EncoderYUV420 : public Encoder {
     if (clipped) {
       const int sub_w = ((W_ + 1) >> 1) - mb_x * 8;
       const int sub_h = ((H_ + 1) >> 1) - mb_y * 8;
-      Convert8To16bClipped(U, v_step_, out + 4 * 64, sub_w, sub_h);
+      Convert8To16bClipped(U, u_step_, out + 4 * 64, sub_w, sub_h);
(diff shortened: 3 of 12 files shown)


src/dichotomy.cc
hasinoff: **⚠️ Bug / MSAN Use-of-Uninitialized-Value on Early Search Failure**

When running `TEST(AllocationFailure)` under MemorySanitizer (MSAN), `FinalizeQuantMatrix` triggers a use-of-uninitialized-value warning.

**Root cause:**
1. `uint8_t opt_quants[2][64];` is allocated on the stack uninitialized at the start of `LoopScan()`.
2. If an allocation fails on pass 0 (`p == 0`) inside `StoreRunLevels()`, `if (!ok_) break;` exits the dichotomy loop before `CopyQuantMatrix(quants_[c].quant_, opt_quants[c])` is ever executed.
3. Execution falls through to `SetQuantMatrices(opt_quants)` and `FinalizeQuantMatrix(&quants_[c], q_bias_)`, which evaluates `if (q->quant_[i] < q->min_quant_[i])` on garbage stack memory from `opt_quants`.
4. Additionally, `StoreRunLevels()` and `StoreOptimalHuffmanTables()` on lines 178–183 were being called even when `!ok_`.

**Suggested fix:** Guard the post-search matrix transfer, Huffman optimization, and bitstream generation with `if (ok_)`:

```diff
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -168,24 +168,24 @@ void Encoder::LoopScan() {
     if (search_hook_->Update(result)) break;
   }
-  // transfer back the final matrices
-  SetQuantMatrices(opt_quants);
-  for (int c = 0; c < 2; ++c) FinalizeQuantMatrix(&quants_[c], q_bias_);
-
-  // return informative values to the user
-  search_hook_->q = best_q;
-  search_hook_->value = best_result;
-
-  // optimize Huffman table now, if we haven't already during the search
-  if (!search_hook_->for_size || !last_is_best) {
-    StoreRunLevels(base_coeffs);
-    if (ok_ && optimize_size_) {
-      StoreOptimalHuffmanTables(nb_mbs, base_coeffs);
-    }
-  }
-
-  // finish bitstream
   if (ok_) {
+    // transfer back the final matrices
+    SetQuantMatrices(opt_quants);
+    for (int c = 0; c < 2; ++c) FinalizeQuantMatrix(&quants_[c], q_bias_);
+
+    // return informative values to the user
+    search_hook_->q = best_q;
+    search_hook_->value = best_result;
+
+    // optimize Huffman table now, if we haven't already during the search
+    if (!search_hook_->for_size || !last_is_best) {
+      StoreRunLevels(base_coeffs);
+      if (ok_ && optimize_size_) {
+        StoreOptimalHuffmanTables(nb_mbs, base_coeffs);
+      }
+    }
+
     WriteDQT();
     WriteSOF();
     WriteDHT();
```

hasinoff, line 20: Add direct `#include <assert.h>`

src/dichotomy.cc
```diff
@@ -213,7 +215,7 @@ size_t Encoder::HeaderSize() const {
       size += (xmp_.size() / 65458 + 1) * 40;
     }
   }
-  size += 2 * 65 + 2 + 2;         // DQT
+  size += (nb_comps_ == 1 ? 1 : 2) * 65 + 2 + 2;  // DQT
```
hasinoff, line 223: Commit 10 fixes the DQT table count from 2 to 1 in `HeaderSize()` for grayscale images (`nb_comps_ == 1`). It would be great to add a test in `tests/unit_test.cc` verifying that `TARGET_SIZE` rate control on `SJPEG_YUV_400` converges accurately to the requested byte target.

src/enc.cc
hasinoff: **Integer Overflow on Valid 16-bit Max Dimensions ($65535 \times 65535$)**

While Commit 6 correctly rejects dimensions $> 65535$, valid large dimensions (e.g. $65535 \times 65535$) cause 32-bit signed integer overflow in intermediate calculations:

1. In `EncoderSharp420`:
   `W * H` evaluates as signed 32-bit `int` ($65535^2 = 4{,}294{,}836{,}225 > \text{INTMAX}$), overflowing to negative values before converting to `size_t`.
   ```cpp
   // Sizing buffer safely:
   const size_t y_size = static_cast<size_t>(W) * H;
   const size_t uv_size = static_cast<size_t>(uv_w) * uv_h;
   yuv_memory_ = Alloc<uint8_t>(y_size + 2 * uv_size);
   if (ok_) {
     y_ = yuv_memory_;
     y_step_ = W;
     u_ = yuv_memory_ + y_size;
     v_ = u_ + uv_size;
     ...
   }
   ```
2. In `SjpegEncode()`:
   ```cpp
   MemorySink sink(static_cast<size_t>(width) * height / 4);
   ```

hasinoff, line 27: - Add `#include "bit_writer.h"`: `enc.cc` directly constructs and uses `MemorySink sink(...)`, which is declared in `bit_writer.h` (and not re-exported by `sjpegi.h`).
- Add `#include <algorithm>` for `std::min` / `std::max`.

src/jpeg_tools.cc
hasinoff, line 23: **Heads-up on Grayscale Detection & Direct Includes:**

1. **Grayscale Detection (Commit 9)**: Fixing the `gray_count` ratio calculation in `SjpegRiskiness` works nicely and correctly detects small/flat images as grayscale (`SJPEG_YUV_400`). As a side effect, bitstreams for small flat images drop below 300 bytes (1 component, 1 DQT). Downstream test harnesses and fuzzers corrupting JPEG header bytes should ensure they bound random offset indices to `std::min(jpeg.size(), max_bytes)` rather than indexing against raw RGB buffer dimensions.
2. **Direct Includes**: Add direct `#include <algorithm>` for `std::min` and `#include <stdlib.h>` for `abs`.

src/yuv_convert.cc
```diff
@@ -111,9 +112,9 @@ static uint32_t kLinearToGammaTab[GAMMA_TABLE_SIZE + 2];
 static uint32_t kGammaToLinearTab[MAX_Y_T + 1];   // size scales with Y_FIX
 
 static void InitGammaTablesF() {
-  static bool done = false;
+  static std::once_flag once;
   assert(2 * GAMMA_TO_LINEAR_BITS < 32);  // we use uint32_t intermediate values
-  if (!done) {
+  std::call_once(once, []() {
```
hasinoff, line 117: **Test Coverage Suggestion (Commit 11)**

Using `static std::once_flag` and `std::call_once` ensures thread safety for the lazy lookup tables. It would be helpful to add a test in `tests/unit_test.cc` that spawns multiple concurrent `std::thread` workers simultaneously calling `SjpegCompress` with `SJPEG_YUV_SHARP` to verify race-free execution under ThreadSanitizer.

tests/unit_test.cc
hasinoff, line 1: **Empty Vector UB, Direct Includes & Test Additions:**
1. **Direct Includes**:
   - Add `#include <algorithm>` for `std::min` / `std::max`.
2. **Empty Vector UB**: In `EncodeRGB()` and `InvalidArguments()`, prefer `vec.data()` over `&vec[0]` to avoid undefined behavior on empty vectors in C++11.
3. **Negative Strides & Odd Dimensions**: Consider adding test cases with negative strides (`stride < 0` for vertically flipped input buffers) in `TEST(EncodeYUV420Strides)` and `TEST(EncodeNV)`, as well as odd dimensions (e.g. $17 \times 13$) to verify sub-sampling boundary clipping.

hasinoff (commented):
Thanks for putting together this extensive cleanup!

We ran the entire 14-commit series through extensive testing, including our full integration test suites, high-iteration fuzzing (**5,000,000 iterations** under ASAN `libFuzzer`), and multi-sanitizer stress testing (**ASAN**, **MSAN**, **UBSAN**, **TSAN**).

All core functionality compiles cleanly and tests pass with zero regressions. In our inline comments below, we have noted:
1. **A MemorySanitizer (MSAN) uninitialized stack read defect** in `src/dichotomy.cc` during allocation failure handling (`TEST(AllocationFailure)`), along with a patch.
2. **Integer overflow edge cases** in `src/enc.cc` when evaluating valid max dimensions ($65535 \times 65535$).
3. **Missing direct includes** (`bit_writer.h` for `MemorySink`, `<assert.h>`, `<algorithm>`).
4. **Suggested test additions** in `tests/unit_test.cc` to cover single-component grayscale rate control, multithreaded `std::call_once` lazy initialization, and negative strides.

src/dichotomy.cc
skal-aa: Fixed in 608a89c, using your patch -- except that the bitstream writes keep their own inner `if (ok_)`: the `StoreRunLevels()` call inside the guarded block can still clear `ok_` through `CheckBuffers()`.

skal-aa: Added.

src/dichotomy.cc
skal-aa: Added `TEST(TargetSize)`: `SJPEG_YUV_400` converges within 1.4%, same as `SJPEG_YUV_420`, against 4.8-8.4% before the fix.

src/enc.cc
skal-aa: Fixed: the plane sizes are computed as `size_t`, and all six `width * height / 4` hints are cast now (only two were). The sharp converter is still int-indexed internally, so 65535x65535 needs a separate pass.

skal-aa: `bit_writer.h` added, along with `<string.h>` and `<string>`. `<algorithm>` is not needed: `enc.cc` has no `std::min` or `std::max`, and the `std::abs` it does use comes from `<cstdlib>`.

src/jpeg_tools.cc
skal-aa: Noted on the fuzzer bounds, thanks. Added `<utility>` for the `std::swap` at line 193; `<algorithm>` and `<stdlib.h>` are not needed, as this file has no `min`, `max` or `abs`.

src/yuv_convert.cc
skal-aa: Added `TEST(Threads)`: eight concurrent `SJPEG_YUV_SHARP` encodes, which TSAN flags as a race if the `call_once` is reverted.

tests/unit_test.cc
skal-aa: Switched to `.data()`, and added `TEST(NegativeStrides)` (RGB / YUV420 / NV12 against the flipped picture) plus 17x13 to the U/V stride tests. `<algorithm>` is not needed here either.

src/dichotomy.cc
skal65535: Fixed in 608a89c, using your patch -- except that the bitstream writes keep their own inner `if (ok_)`: the `StoreRunLevels()` call inside the guarded block can still clear `ok_` through `CheckBuffers()`.

skal65535: Added.

src/dichotomy.cc
skal65535: Added `TEST(TargetSize)`: `SJPEG_YUV_400` converges within 1.4%, same as `SJPEG_YUV_420`, against 4.8-8.4% before the fix.

src/enc.cc
skal65535: Fixed: the plane sizes are computed as `size_t`, and all six `width * height / 4` hints are cast now (only two were). The sharp converter is still int-indexed internally, so 65535x65535 needs a separate pass.

skal65535: `bit_writer.h` added, along with `<string.h>` and `<string>`. `<algorithm>` is not needed: `enc.cc` has no `std::min` or `std::max`, and the `std::abs` it does use comes from `<cstdlib>`.

src/jpeg_tools.cc
skal65535: Noted on the fuzzer bounds, thanks. Added `<utility>` for the `std::swap` at line 193; `<algorithm>` and `<stdlib.h>` are not needed, as this file has no `min`, `max` or `abs`.

src/yuv_convert.cc
skal65535: Added `TEST(Threads)`: eight concurrent `SJPEG_YUV_SHARP` encodes, which TSAN flags as a race if the `call_once` is reverted.

tests/unit_test.cc
skal65535: Switched to `.data()`, and added `TEST(NegativeStrides)` (RGB / YUV420 / NV12 against the flipped picture) plus 17x13 to the U/V stride tests. `<algorithm>` is not needed here either.

skal65535:
@hasinoff , i fixed all your comments , thanks for the testing! PTAL.

hasinoff:
LGTM

src/jpeg_tools.cc
```diff
@@ -31,27 +32,28 @@ namespace {
...
-  const uint8_t* const end = src + size - 8;   // 8 bytes of safety, for marker
-  src += 2;   // skip M_SOI
-  for (; src < end && *src != 0xff; ++src) { /* search first 0xff marker */ }
-  while (src < end) {
-    const uint32_t marker = static_cast<uint32_t>((src[0] << 8) | src[1]);
-    if (marker == M_SOF0 || marker == M_SOF1) return src;
-    const size_t s = 2 + ((src[2] << 8) | src[3]);
-    src += s;
+// Positions are used instead of pointers, to never build one past the end.
+const uint8_t* GetSOFData(const uint8_t* src, size_t size) {
```
jzern, line 36: Work like this has been transitioning to `std::span`.

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ set(CMAKE_CXX_STANDARD 11)
 # Options for coder / decoder executables.
 option(SJPEG_ENABLE_SIMD "Enable any SIMD optimization." ON)
 option(SJPEG_BUILD_EXAMPLES "Build the sjpeg / vjpeg command line tools." ON)
+option(SJPEG_BUILD_TESTS "Build the unit tests." ON)
```
jzern, line 30: Note `include(CTest)` provides [`BUILD_TESTING`](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html).

skal-aa: ah, right. TODO(later)

src/jpeg_tools.cc
skal-aa: that's C++20 feature, and we're still using C++11 here (probably i bad idea overall)

PR #152, opened 2026-08-18
Split enc.cc + six bit-exact encoder optimizations

skal65535:
Splits `enc.cc`, and add six bit-exact optimizations on top.

On `source3.jpg`:
**-27%** on the default encode and **-47%** on method 7.

On 537 photographs from an Open Images corpus:
**-9.7%** on the default encode and **-27.5%** multi-pass.

### The split

`enc.cc` had grown to 2573 lines doing six unrelated jobs. Every cut lands on a
`////` banner that was already there: no function body is touched, no line is
rewritten, and the encoder measures at 0% either way.

### The optimizations

- **64bit `BitWriter`**: 56 usable bits, so about five symbols pile up per
  flush, and the common case — no `0xff` among the pending bytes — becomes one
  8-byte store instead of a byte loop. DC and AC merge their code and suffix
  into a single `PutBits()`.
- **64bit `BitCounter`**: the same for the multi-pass size estimator. Its flush
  counts the escapes with one popcount rather than writing them.
- **Chunked commit**: the sink was called once per macroblock, a `resize()` per
  MCU for `Sink<std::string>`. The writer now serves those out of a slab.
- **Fused entropy statistics**: the size search walked the run/levels twice.
  The second walk is gone, and `StoreOptimalHuffmanTables()` with it.
- **Trellis prune**: `SearchBestPrev()` is O(n²) with n up to 127. The bound is
  monotone walking back towards the sink, so the search can stop rather than
  skip — an exact early-out, not an approximation.
- **Integer riskiness accumulation**: `SjpegRiskiness()` sweeps the
  whole bitmap. Convert the accumulation with 'double' to 'int64_t' (faster).
  Safe replacement, since 'score' is at most 3 × 255, and the sums are < 2^38.

### Measured

M5 Pro (clang 21), Raspberry Pi 5 / Cortex-A76 (gcc 14), Haswell (gcc 15), on
`source3.jpg`. Each of the first six rows is that change removed on its own, so
they do not add up to the totals.

| | M5 | RPi 5 | Haswell |
|---|---:|---:|---:|
| 64bit `BitWriter` (m0 q75) | -31.0% | -18.4% | -16.1% |
| 64bit `BitCounter` (m4, 6 passes) | -36.6% | -21.9% | -24.8% |
| chunked commit (m4 q75) | -8.3% | -3.2% | noise |
| fused entropy stats (m4, 6 passes) | -5.3% | -1.7% | -3.9% |
| trellis prune (m7 q75) | -43.5% | -47.5% | -46.4% |
| integer riskiness (`SjpegRiskiness` alone) | -22.7% | — | — |
| **all six**, default (m4 AUTO) | **-27.2%** | -11.5%¹ | -9.2%¹ |
| **all six**, method 7 | **-46.9%** | -49.2%¹ | -47.8%¹ |
| *noise floor, worst row* | *1.8%* | *0.6%* | *3.1%* |

¹ first five only — the sixth has not been run on those two machines yet.

Over 537 Open Images originals, 0.07 to 72.6 MP, median 4.98 MP, M5 medians:

| | source3 | corpus |
|---|---:|---:|
| m4 q75 AUTO (default) | -27.2% | **-9.7%** |
| m4 q75 6 passes | -40.6% | **-27.5%** |
| m7 q75 trellis | -46.9% | **-10.6%** |

Multi-pass carries over, single-pass and method 7 do not. The trellis early-out
fires less on dense photographic texture, and `AUTO` picks `sharp420` for 297 of
the 537, whose conversion pass dilutes the rest: those gain -7.2%, against
-19.7% for the 221 on plain 420.

### Note for review

SJPEG_FORCE_32BIT: new build-time switch to force 32bit writer/counter.
Useful for tests on non-64bit platforms.

Future work subsampling`SjpegRiskiness`, subsampling the histogram pass,
and computing the size from the frequency table.

Please rebase-merge rather than squash, so the optimizations stay bisectable.


src/api.cc
```diff
@@ -0,0 +1,307 @@
...
+
+#include <assert.h>
+#include <stdint.h>
+
+#include <cstdlib>
+#include <string>
+
+#include "sjpegi.h"
+
+using namespace sjpeg;
```
jzern: Is this needed? Most everything is within this namespace starting at line 72.

src/bit_writer.h
```diff
@@ -28,6 +28,18 @@
...
 
+// Bit writer and counter below have a 64bit and a 32bit implementation, picked
+// by target register width: a 32bit target would emulate the accumulator in a
+// register pair, which is untested here. Both emit the same bytes.
+// SJPEG_FORCE_32BIT picks the 32bit one on a 64bit host, so it can be built
+// and compared anywhere.
+
+#if !defined(SJPEG_FORCE_32BIT) &&                  \
+    (defined(__LP64__) || defined(_WIN64) ||        \
+     (defined(UINTPTR_MAX) && UINTPTR_MAX > 0xffffffffu))
```
jzern: This last one might be sufficient given the use of `<stdint.h>` (unconfirmed).

src/bit_writer.h
```diff
@@ -77,6 +89,51 @@ template<class T> class Sink : public ByteSink {
...
 
+///////////////////////////////////////////////////////////////////////////////
+
+#if defined(SJPEG_HAVE_64BIT)
+
+// Hottest members below are called once per coefficient. Left to its own cost
+// model, the compiler keeps some of them out of line, which is expensive at
+// that frequency, so we insist. Undefined at end of this header.
+#if defined(__GNUC__) || defined(__clang__)
+#define SJPEG_INLINE inline __attribute__((always_inline))
```
jzern, line 103: `[[gnu::always_inline]]` is an option with C++11, but I don't think there's anything gained.

src/bit_writer.h
```diff
@@ -77,6 +89,51 @@ template<class T> class Sink : public ByteSink {
...
+#endif
+}
+
+// Accumulator holds the bytes to emit in its top bytes, so storing it whole
+// means storing it big-endian: a swap on a little-endian host, nothing on a
+// big-endian one. Getting it wrong reverses every group of eight silently.
+static inline uint64_t HToBE64(uint64_t x) {
+#if (defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) && \
+     __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) ||                  \
+    defined(__BIG_ENDIAN__)
```
jzern: Looks like abseil and ffmpeg are only checking `__BYTE_ORDER__` these days.

jzern (commented):
Only partial review through the first few files. Running some tests locally.

src/encoders.cc
```diff
@@ -0,0 +1,577 @@
...
+//
+//  Per-colorspace Encoder sub-classes, and their factory
+//
+// Author: Skal (pascal.massimino@gmail.com)
+
+#include <assert.h>
+#include <stdint.h>
+#include <string.h>
+
+#include <new>
```
jzern, line 24: Add `<cstdlib>` for `std::abs`.

src/api.cc
skal65535: Needed: `MemorySink`/`Encoder`/`EncoderFactory` are used unqualified in the plain-C entry points above the `namespace sjpeg {` block. Added a comment saying so in 8bbc420.

src/bit_writer.h
skal65535: Confirmed, `<stdint.h>` is included unconditionally so `UINTPTR_MAX` alone is sufficient. Simplified in 8bbc420.

src/bit_writer.h
skal65535: Agreed, no gain -- leaving the `__attribute__` form.

src/bit_writer.h
skal65535: Simplified to just `__BYTE_ORDER__` in 8bbc420.

src/encoders.cc
skal65535: Added in 8bbc420.

src/bit_writer.h
```diff
@@ -280,11 +280,60 @@ struct BitCounter {
...
+  size_t Size() { Flush(); return size_; }
+
+ private:
+  // Number of bytes equal to 0xff among the 'nb' most significant ones.
+  static int CountFF(uint64_t v, int nb) {
+    // Usual haszero() trick is not usable as-is here: it only answers "is there
+    // one", because a zero byte borrows into its neighbour and marks it too.
+    // Masking to 0x7f before the add keeps carries local, which makes the
+    // result countable.
+    const uint64_t x = ~v;   // 0xff bytes are now zero bytes
```
hasinoff: (Optional) A little code golf to save instructions:
  const uint64_t y = (v & 0x7f7f7f7f7f7f7f7full) + 0x0101010101010101ull;
  const uint64_t z = (y & v & 0x8080808080808080ull) >> (64 - 8 * nb);

No measurable perf impact. If you're willing to add a LUT you can squeeze a little performance, but not much impact e2e:
    // A compile-time lookup table isolating the Most Significant Bit (MSB) 
    // of the 'nb' most significant bytes. 
    static constexpr uint64_t MASKS_80[9] = {
        0x0000000000000000ULL, // nb = 0
        0x8000000000000000ULL, // nb = 1
        0x8080000000000000ULL, // nb = 2
        0x8080800000000000ULL, // nb = 3
        0x8080808000000000ULL, // nb = 4
        0x8080808080000000ULL, // nb = 5
        0x8080808080800000ULL, // nb = 6
        0x8080808080808000ULL, // nb = 7
        0x8080808080808080ULL  // nb = 8
    };

skal65535: @hasinoff  your code suggestion is indeed a wash, speed-wise (micro-benchmarked on an M5, -O3).
But I incorporated it in, since it's less instructions :)

jzern:
Bit exact in my testing. Some rough numbers show ~1.09-1.24 times faster depending on the test set.

src/api.cc
```diff
@@ -0,0 +1,308 @@
...
+
+#include <assert.h>
+#include <stdint.h>
+
+#include <cstdlib>
+#include <string>
+
+#include "sjpegi.h"
+
+using namespace sjpeg;  // for the plain-C entry points below, which sit
```
jzern: Either way works, but the fully qualified form is no all that long...

src/bit_writer.h
```diff
@@ -77,6 +88,49 @@ template<class T> class Sink : public ByteSink {
...
+// returns true if any byte of 'x' is 0xff
+static inline bool HasFF(uint64_t x) {
+  const uint64_t y = ~x;    // 0xff bytes are now zero bytes
+  return ((y - 0x0101010101010101ull) & ~y & 0x8080808080808080ull) != 0;
+}
+
+static inline uint64_t BSwap64(uint64_t x) {
+#if defined(__GNUC__) || defined(__clang__)
+  return __builtin_bswap64(x);
+#else
```
jzern, line 121: From [webp](https://chromium.googlesource.com/webm/libwebp/+/refs/tags/v1.6.0/src/utils/endian_inl_utils.h#84):

```
#elif defined(_MSC_VER)
  return (uint64_t)_byteswap_uint64(x);
```

src/bit_writer.h
```diff
@@ -159,15 +287,65 @@ struct BitCounter {
...
+  static int CountFF(uint64_t v, int nb) {
+    // Masking to 0x7f before the add keeps carries local: a byte's low 7 bits
+    // are all-ones (0x7f or 0xff) iff adding 1 sets bit 7. ANDing that back
+    // against the byte's own original bit 7 keeps only the 0xff case, since
+    // 0x7f has its top bit clear.
+    const uint64_t y = (v & 0x7f7f7f7f7f7f7f7full) + 0x0101010101010101ull;
+    uint64_t z = (y & v & 0x8080808080808080ull);
+    z >>= 64 - 8 * nb;                 // keep the top 'nb' bytes
+#if defined(__GNUC__) || defined(__clang__)
+    return __builtin_popcountll(z);
```
jzern, line 338: `<intrin.h>` on `_MSC_VER` has `__popcnt64`: https://learn.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64?view=msvc-170

src/encoders.cc
```diff
@@ -0,0 +1,578 @@
...
+             ByteSink* const sink, PixelFormat fmt = kRGBInput,
+             MemoryManager* const memory = nullptr)
+      : Encoder(SJPEG_YUV_420, W, H, sink, memory), rgb_(rgb), step_(step) {
+    ok_ = (rgb != nullptr);
+    if (fmt != kRGBInput) {
+      pix_step_ = 4;
+      get_yuv_block_ = GetBlockFunc(yuv_mode_, fmt);
+    }
+  }
+  virtual ~Encoder420() {}
```
jzern: Could be `override` now, but this is a preexisting condition.

src/encoders.cc
```diff
@@ -0,0 +1,578 @@
...
+    if (clipped) {
+      rgb = GetReplicatedSamples(rgb, step_,
+                                 W_ - mb_x * 8, H_ - mb_y * 8, 8, 8);
+      step = pix_step_ * 8;
+    }
+    get_yuv_block_(rgb, step, out);
+  }
+
+ protected:
+  const uint8_t* rgb_;
```
jzern: Can this declaration match 420/444 (line 186 and 219)?

src/encoders.cc
```diff
@@ -0,0 +1,578 @@
...
+    const uint8_t* data = gray_ + (mb_x + mb_y * step_) * 8;
+    if (clipped) {
+      Convert8To16bClipped(data, step_, out, W_ - mb_x * 8, H_ - mb_y * 8);
+    } else {
+      Convert8To16b(data, step_, out);
+    }
+  }
+
+ protected:
+  const uint8_t* gray_;
```
jzern: Same here

jzern (approved):

src/api.cc
skal65535: indeed, done

src/bit_writer.h
skal65535: done

src/bit_writer.h
skal65535: done below

src/encoders.cc
skal65535: yep, fix later

src/encoders.cc
skal65535: done

src/encoders.cc
skal65535: done

skal65535:
> Bit exact in my testing. Some rough numbers show ~1.09-1.24 times faster depending on the test set.

Nice. I think i fixed all the comment, so should be good to go. Please submit when you think it ok...

jzern:
Lots's of old conversion related warnings from visual studio (double->float, size_t->int, etc.), but builds. Probably could look into a workflow for that at some point. libavif has an example for both cl.exe and clang-cl.

skal65535:
clang's `-Wshorten-64-to-32` finds 194 warnings for conversions. Should be easy to auto-fix the most annoying...

PR #156, opened 2026-08-20
dichotomy: derive ComputeSize() from the frequency table

skal65535:
* StoreRunLevels() already gathers freq_ac_/freq_dc_ whenever optimize_size_ is set
=> which is always true at LoopScan()'s only call site.

* but ComputeSize() nevertheless still re-walked every run/level through a BitCounter,
  instead of using the frequency distribution already collected.

Using EntropySize() is faster (measured ~13% on multi-pass encodes) but a bit approximate
because it doesn't count the 0xff byte-stuffing (BitCounter does incorporate it).
Observed diff on Open Images corpus is 0.92% vs 0.90% before, acceptable.

The expected underestimation (neglecting 0xff-stuffing) is 1/256 (assuming uniform byte distribution, only 0xff produce two bytes instead of one), that's 0.39%. Not worth correcting the estimate with a wild-guess, it's under the convergence precision anyway.

### DIFF
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -282,9 +282,18 @@ void Encoder::BlocksSize(int nb_mbs, const DCTCoeffs* coeffs,
 float Encoder::ComputeSize(const DCTCoeffs* coeffs) {
   InitCodes(false);
   size_t size = HeaderSize();
-  BitCounter bc;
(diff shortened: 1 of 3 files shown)


jzern (approved):

skal65535:
this PR is based on skal/enc-perf, hence waiting for it to be submitted

PR #158, opened 2026-08-21
clang tidy

jzern:
- **add missing / delete unused includes**
- **sjpegi.h: remove const qualification in decls**

### DIFF
--- a/src/sjpegi.h
+++ b/src/sjpegi.h
@@ -531,17 +531,17 @@ struct Encoder {
 // Returns the Encoder sub-class matching 'yuv_mode' (resolving SJPEG_YUV_AUTO
 // along the way), or null if it could not be constructed.
 extern Encoder* EncoderFactory(const uint8_t* rgb, int W, int H, int stride,
-                               SjpegYUVMode yuv_mode, ByteSink* const sink,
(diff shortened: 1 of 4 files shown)


jzern:
This can be rebased and merged if both commits look good to you.

src/sjpegi.h
```diff
@@ -527,17 +531,17 @@ struct Encoder {
 // Returns the Encoder sub-class matching 'yuv_mode' (resolving SJPEG_YUV_AUTO
 // along the way), or null if it could not be constructed.
 extern Encoder* EncoderFactory(const uint8_t* rgb, int W, int H, int stride,
-                               SjpegYUVMode yuv_mode, ByteSink* const sink,
+                               SjpegYUVMode yuv_mode, ByteSink* sink,
```
skal65535, line 534: so we're using the convention of keeping the extra 'const' in the .cc but removing it in the .h for clarity?

skal65535 (approved):

src/sjpegi.h
jzern: Yes, it's this one: https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html.

We only need them both to match when dealing with .c files, mostly to appease Microsoft's compiler.

PR #161, opened 2026-08-21
Add progressive JPEG encoding

### LINKED ISSUE #124: progressive coding
jpegrescan can sometimes shrink the size further. There should be an option to have progressive-coded output.

Starting a branch 'prog' for that.

skal65535:
Adds progressive JPEG encoding

the code can be disabled by defining SJPEG_NO_PROGRESSIVE

---

- 1 interleaved DC scan, followed by 1 or 2 non-interleaved AC scans/component,
- luma/chroma scans are split at a configurable spectral position (`Ss`/`Se`),
- use EOBn run-length coding across empty blocks (<- that's the real win!)

- No successive approximation (`Ah`/`Al` are always 0). Not worth the code complexity.

- Off by default (default sequential coding is bit-exact and unchanged in speed)
- Can also be stripped from the build entirely with `-DSJPEG_HAVE_PROGRESSIVE=0`
  (CMake option or Makefile/Android.mk variable) -- `src/prog.cc` is excluded
  from the build outright in that case, not just its symbols.

- `-prog x[,y]` sets the luma/chroma split points directly
- `-progressive` is a shortcut for `-prog 2,8`
- this default split values were jpegtran-cross-validated with a corpus sweep
- `-progressive`: -2.5% mean / -3.4% total-corpus size, winning on 95% of cases

---

- `src/prog.cc`: plane allocation and the DC/AC scan drivers
  (`EncodeProgressive`, `EncodeProgAC`).
- `src/entropy.cc`: windowed AC coefficient coding and EOBn run-length
  coding for the progressive scans.
- `src/headers.cc`: `WriteOneDHT`/`WriteProgSOS` for per-scan tables and
  explicit-component SOS markers.
- `examples/sjpeg.cc`: `-prog`/`-progressive` CLI options.
-  fixed and updated `man/sjpeg.1`, `man/vjpeg.1` (was gone stale)

Two real bugs found by fuzzing CLI option combinations against exotic inputs
under ASan/UBSan, both fixed: progressive + `-trellis` crashed (a missing
`InitCodes(true)` call left `ac_codes_[]` uninitialized), and a malformed PPM
header could overflow the size computation in `examples/utils.cc`'s reader.

Bumped version to 0.1.2

Fixes #124


hasinoff:
Didn't look at the code in detail, but I can report it's passing Google's global presubmits and is performance neutral with default settings (as expected).

src/sjpeg.h
hasinoff: Consider bumping version, given new API's

skal65535: done (->0.1.2)

examples/sjpeg.cc
```diff
@@ -160,15 +171,15 @@ int main(int argc, char * argv[]) {
     } else if (!strcmp(argv[c], "-q") && c + 1 < argc) {
       quality = atof(argv[++c]);
       use_reduction = false;
-      if (quality < 0 || quality > 100) {
+      if (!(quality >= 0 && quality <= 100)) {  // catches NaN too
```
jzern, line 174: subtle

examples/utils.cc
```diff
@@ -893,7 +893,9 @@ vector<uint8_t> ReadPPM(const std::string& input,
       max_value > 255) {
     return rgb;
   }
-  const size_t rgb_size = W * H * 3;
+  if (W <= 0 || H <= 0) return rgb;
+  // avoid int overflow
+  const size_t rgb_size = (size_t)W * H * 3;
```
jzern: What about a 32-bit build?

man/sjpeg.1
```diff
@@ -31,22 +31,104 @@ is 75.
...
 option.
 .TP
+.B \-estimate
+If the source is a JPEG file, print the estimated quality and exit.
+.TP
+.B \-i, \-info
+Print some information about the input file and exit.
+.TP
+.B \-quiet
+Disable information message (error messages are still printed).
```
jzern: Maybe 'informational messages'? To match the 'error messages' part.

man/sjpeg.1
```diff
@@ -31,22 +31,104 @@ is 75.
 A small factor produces a smaller file
 with lower quality. Best quality is achieved by using a value of 100.
 If the input source is a JPEG file, the compression factor will be limited
-to never exceed the source's quality unless the -no_limit option is used.
+to never exceed the source's quality unless the \-no_limit option is used.
```
jzern: Could be `\fB\-no_limit\fP`

man/sjpeg.1
```diff
@@ -31,22 +31,104 @@ is 75.
...
+.TP
+.B \-progressive
+Use progressive encoding, with the recommended spectral split points.
+Shortcut for \fB\-prog 2,8\fP.
+.TP
+.BI \-prog " int[,int]
+Enable progressive encoding, and set the low/high frequency split point
+for luma to the first value (1..63), and optionally for chroma to the
+second value (defaults to 8 if omitted). Disabled unless this option is
+used; \fB2,8\fP is a well-tested starting point, see \-progressive above.
```
jzern: Could be `\fB-progressive\fP`

src/enc.cc
```diff
@@ -128,6 +134,23 @@ void Encoder::SetCompressionMethod(int method) {
...
+void Encoder::SetProgressive(int luma_split, int chroma_split) {
+#if !defined(SJPEG_NO_PROGRESSIVE)
+  luma_split = (luma_split < 1) ? 1 : (luma_split > 64) ? 64 : luma_split;
+  chroma_split = (chroma_split < 1) ? 1 : (chroma_split > 64) ? 64
+                                                              : chroma_split;
+  progressive_ = (luma_split < 64);
+  prog_luma_split_ = luma_split;
+  prog_chroma_split_ = chroma_split;
+  // use_extra_memory_/reuse_run_levels_ are baseline-only; EncodeProgressive()
+  // has its own storage and needs neither.
```
jzern: neither? Did you mean to say 'EncodeProgressive() has its own storage and needs.'? or it doesn't?

tests/unit_test.cc
```diff
@@ -645,6 +686,135 @@ TEST(QuantMatrix) {
...
+  {
+    sjpeg::EncoderParam param(75.f);
+    param.progressive_luma_split = 2;
+    std::string out;
+    CHECK(EncodeRGB(rgb, kWidth, kHeight, param, &out));
+
+    sjpeg::EncoderParam param2(75.f);
+    param2.progressive_luma_split = 2;
+    param2.target_mode = sjpeg::EncoderParam::TARGET_SIZE;
+    param2.target_value = static_cast<float>(out.size());
```
jzern: Did you want a C-style cast here for consistency elsewhere?

CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ set(CMAKE_CXX_STANDARD 11)
 option(SJPEG_ENABLE_SIMD "Enable any SIMD optimization." ON)
 option(SJPEG_BUILD_EXAMPLES "Build the sjpeg / vjpeg command line tools." ON)
 option(SJPEG_BUILD_TESTS "Build the unit tests." ON)
+option(SJPEG_HAVE_PROGRESSIVE "Build with progressive JPEG encoding support." ON)
```
jzern: `ENABLE` to match `SJPEG_ENABLE_SIMD`?

jzern (approved):

man/sjpeg.1
skal65535: done

man/sjpeg.1
skal65535: done

tests/unit_test.cc
skal65535: indeed

CMakeLists.txt
skal65535: good call, done

src/enc.cc
skal65535: 'neither of them'

man/sjpeg.1
skal65535: done

examples/utils.cc
skal65535: good catch! Used uint64_t and then size_t

jzern (approved):

PR #162, opened 2026-08-22
More bit-exact sjpeg optimizations

hasinoff:
Optimizes several preprocessing and analysis routines in sjpeg. Adds the SJPEG_UNROLL(n) cross-compiler directive macro in src/sjpegi.h (supporting Clang and GCC) and applies targeted unrolling and memory optimizations.

- SjpegRiskiness (src/jpeg_tools.cc): 4x unrolling with SJPEG_UNROLL(4) and branchless neutral-chroma range check, preserving the sparse score branch. Improves kernel throughput by +10.6% on GCC while remaining neutral on Clang (~501 MP/s, cache-bound), and improves end-to-end AUTO mode encoding throughput by +3.7% (Clang) / +2.7% (GCC).

- PreprocessARGB scratch arena (src/yuv_convert.cc): Consolidates 7 separate dynamic std::vector heap allocations into a single contiguous scratch arena buffer (allocated without redundant zero-initialization) and adds allocation failure error handling. Eliminates ~14 MB of per-frame memory clearing and allocator lock overhead, improving Sharp YUV end-to-end throughput by +1.2-1.6% (Clang) / +1.4-1.8% (GCC).

- Histogram generation (src/histogram.cc): Uses SSSE3 _mm_abs_epi16 (pabsw) in StoreHistoSSE2 when available, reducing instruction count for block histogram accumulation and improving kernel throughput by +1.8% on Clang and +1.2-7.8% on GCC (+0.3-0.7% end-to-end impact in adaptive quantization modes).

- Quality estimation (src/jpeg_tools.cc): 4x unrolling in SjpegEstimateQuality difference loop, improving quality estimation throughput by +1.9-3.7% (Clang) / +3.0-4.3% (GCC) during header/table inspection (0.0% end-to-end impact on encoding as it is not part of the encode path).

All changes are 100% bit-exact against baseline and pass all unit tests and sanitizers (ASAN, MSAN, UBSAN, TSAN).

### DIFF
--- a/src/histogram.cc
+++ b/src/histogram.cc
@@ -60,8 +60,12 @@ void StoreHistoSSE2(const int16_t in[64], Histo* const histos, int nb_blocks) {
     for (int i = 0; i < 64; i += 8) {
       const __m128i A =
           _mm_loadu_si128(reinterpret_cast<const __m128i*>(in + i));
+#if defined(SJPEG_USE_SSSE3)
+      const __m128i C = _mm_abs_epi16(A);
+#else
       const __m128i B = _mm_srai_epi16(A, 15);                  // sign extract
       const __m128i C = _mm_sub_epi16(_mm_xor_si128(A, B), B);  // abs(A)
+#endif
       const __m128i D = _mm_srli_epi16(C, HSHIFT);              // >>= HSHIFT
       const __m128i E = _mm_min_epi16(D, kMaxHisto);
       _mm_storeu_si128(reinterpret_cast<__m128i*>(tmp + i), E);
--- a/src/jpeg_tools.cc
+++ b/src/jpeg_tools.cc
@@ -135,7 +135,7 @@ void SjpegQuantMatrix(float quality, bool for_chroma, uint8_t matrix[64]) {
   const float q_factor = sjpeg::GetQFactor(quality) / 100.f;
   const uint8_t* const matrix0 = sjpeg::kDefaultMatrices[for_chroma];
   for (int i = 0; i < 64; ++i) {
-    const int v = static_cast<int>(matrix0[i] * q_factor + .5f);
+    const int v = (int)(matrix0[i] * q_factor + .5f);
     matrix[i] = (v < 1) ? 1u : (v > 255) ? 255u : v;
   }
 }
@@ -151,6 +151,7 @@ float SjpegEstimateQuality(const uint8_t matrix[64], bool for_chroma) {
     uint8_t m[64];
     SjpegQuantMatrix(quality, for_chroma, m);
     float score = 0;
+    SJPEG_UNROLL(4)
     for (size_t i = 0; i < 64; ++i) {
       const float diff = m[i] - matrix[i];
       score += diff * diff;
@@ -197,10 +198,11 @@ SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
     rgb += stride;
     std::swap(row1, row2);
     cvrt_func(rgb, width, &row2[0]);  // this is the row below
+    SJPEG_UNROLL(4)
     for (int i = 0; i < width - 1; ++i) {
       const int idx0 = row1[i + 0];
       const int idx1 = row1[i + 1];
-      const int idx2 = row2[i];
+      const int idx2 = row2[i + 0];
       const int score = sjpeg::kSharpnessScore[idx0 + kRGB3 * idx1]
                       + sjpeg::kSharpnessScore[idx0 + kRGB3 * idx2]
                       + sjpeg::kSharpnessScore[idx1 + kRGB3 * idx2];
@@ -208,7 +210,7 @@ SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
         score_sum += score;
         score_num += 1;
       }
-      gray_num += (idx0 >= gray_min && idx0 < gray_min + s);
+      gray_num += ((uint32_t)(idx0 - gray_min) < (uint32_t)s);
     }
   }
   const double count = (double)score_num;
--- a/src/yuv_convert.cc
+++ b/src/yuv_convert.cc
@@ -549,13 +549,13 @@ static void ConvertWRGBToYUV(const fixed_y_t* best_y,
                              int width, int height,
                              uint8_t* y_plane,
                              uint8_t* u_plane, uint8_t* v_plane) {
-  const int w = (width + 1) & ~1;
-  const int h = (height + 1) & ~1;
-  const int uv_w = w >> 1;
-  const int uv_h = h >> 1;
-  for (int j = 0; j < height; ++j) {
-    const int off = (j >> 1) * 3 * uv_w;
-    for (int i = 0; i < width; ++i) {
+  const size_t w = ((size_t)width + 1) & ~1ULL;
+  const size_t h = ((size_t)height + 1) & ~1ULL;
+  const size_t uv_w = w >> 1;
+  const size_t uv_h = h >> 1;
+  for (size_t j = 0; j < (size_t)height; ++j) {
+    const size_t off = (j >> 1) * 3 * uv_w;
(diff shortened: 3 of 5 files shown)


src/yuv_convert.cc
```diff
@@ -591,15 +591,32 @@ static void PreprocessARGB(const uint8_t* const rgb,
...
+  ptr += w * h;
+  fixed_y_t* const best_rgb_y = ptr;
+  ptr += w * 2;
+  fixed_t* const best_uv = (fixed_t*)ptr;
+  ptr += uv_w * 3 * uv_h;
+  fixed_t* const target_uv = (fixed_t*)ptr;
+  ptr += uv_w * 3 * uv_h;
+  fixed_t* const best_rgb_uv = (fixed_t*)ptr;
+  ptr += uv_w * 3 * 1;
+  (void)ptr;
```
skal65535: needed?

src/yuv_convert.cc
```diff
@@ -591,15 +591,32 @@ static void PreprocessARGB(const uint8_t* const rgb,
...
-  vector<fixed_t> best_rgb_uv(uv_w * 3 * 1);
-  const uint64_t diff_y_threshold = static_cast<uint64_t>(3.0 * w * h);
+  // Single memory chunk allocation to avoid allocator lock contention and heap
+  // fragmentation
+  const size_t total_elems = (size_t)(w * 3 * 2) + (size_t)(w * h) +
+                             (size_t)(w * h) + (size_t)(w * 2) +
+                             (size_t)(uv_w * 3 * uv_h) +
+                             (size_t)(uv_w * 3 * uv_h) + (size_t)(uv_w * 3 * 1);
+  std::unique_ptr<uint16_t[]> arena(new (std::nothrow) uint16_t[total_elems]);
+  if (arena == nullptr) return;
```
skal65535: this is pre-existing, but we should report back alloc failure here

skal65535: maybe remove the std::nothrow above and let exceptions handle the alloc-failures like vector<> was doing before

hasinoff: Plumbed a bool return value through.

For consistency with all the other allocs (in encoders.cc) I think we should stick with new (std::nothrow).

src/yuv_convert.cc
hasinoff: This pattern advances the ptr past the last chunk for consistency and self-documentation, and then the (void) read avoids the pedantic warning "Value stored to 'ptr' is never read".

I changed the last 2 lines to an assert instead.

src/histogram.cc
```diff
@@ -60,9 +60,13 @@ void StoreHistoSSE2(const int16_t in[64], Histo* const histos, int nb_blocks) {
     for (int i = 0; i < 64; i += 8) {
       const __m128i A =
           _mm_loadu_si128(reinterpret_cast<const __m128i*>(in + i));
+#if defined(SJPEG_USE_SSSE3)
```
skal65535, line 63: Makefile is missing the -mssse3 flag to activate this optim. Similarly  cmake/cpu.make only detects SSE2.

This optim is good (verified) but i'll send an update to the make system later to enable it.

src/jpeg_tools.cc
```diff
@@ -197,18 +198,18 @@ SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
...
-      if (score > kNoiseLevel) {
-        score_sum += score;
-        score_num += 1;
-      }
-      gray_num += (idx0 >= gray_min && idx0 < gray_min + s);
+      const int idx2 = row2[i + 0];
+      const int score = sjpeg::kSharpnessScore[idx0 + kRGB3 * idx1] +
+                        sjpeg::kSharpnessScore[idx0 + kRGB3 * idx2] +
+                        sjpeg::kSharpnessScore[idx1 + kRGB3 * idx2];
+      const bool m = (score > kNoiseLevel);
```
skal65535: this change is killing the perf on x86/gcc... more details later

hasinoff: Oops, the perf numbers reported for this were totally wrong - taking scalar C as a baseline, and the gains were almost totally due to RowIndex SIMD vs. not. I'm doing a more thorough audit of the whole PR.

I can reproduce the regression you're seeing using gcc 15.

What I'm now seeing on my end, in isolation:
- The change here is actually perf neutral on clang
- However, keeping the loop unroll directive but bringing back the branching:
```
      if (score > kNoiseLevel) {
        score_sum += score;
        score_num += 1;
      }
```
improves gcc perf on this kernel +10%, making it match clang, while not affecting clang perf significantly.

src/histogram.cc
hasinoff: The perf impact here is pretty minor e2e, but it would be good if sjpeg could detect SSSE3 and AVX2 support.
There are some bigger opportunities for AVX2.

hasinoff:
Updated the PR:

- Restored branching in SjpegRiskiness (if (score > kNoiseLevel)) while keeping 4x unrolling and unsigned range check, eliminating the GCC performance regression (+10.6% on GCC, neutral on Clang).
- Removed unrolling from AnalyseHisto in src/histogram.cc.
- Updated the description with verified, isolated benchmark numbers for each change across both Clang and GCC, including explicit end-to-end pipeline impacts.

src/histogram.cc
skal65535: added an issue for that, i'll give it a try

src/jpeg_tools.cc
skal65535: re-tested on x86/gcc and arm/clang, it's now ok indeed.

src/yuv_convert.cc
```diff
@@ -591,15 +594,31 @@ static void PreprocessARGB(const uint8_t* const rgb,
...
-  vector<fixed_y_t> best_y(w * h);
-  vector<fixed_y_t> target_y(w * h);
-  vector<fixed_y_t> best_rgb_y(w * 2);
-  vector<fixed_t> best_uv(uv_w * 3 * uv_h);
-  vector<fixed_t> target_uv(uv_w * 3 * uv_h);
-  vector<fixed_t> best_rgb_uv(uv_w * 3 * 1);
-  const uint64_t diff_y_threshold = static_cast<uint64_t>(3.0 * w * h);
+  // Single memory chunk allocation to avoid allocator lock contention and heap
+  // fragmentation
+  const size_t total_elems = (size_t)(w * 3 * 2) + (size_t)(w * h) +
```
skal65535: so there's a pre-existing security hazard here: 'w * h' can overflow. We should use `(size_t)w * h` instead of `(size_t)(w * h)`.

I think the simplest is to make 'w' and 'h' variables be 'size_t' type. Or switch to '(size_t)w * h'.

hasinoff: Done, also addressed in some related functions.

skal65535:
excellent! This settles it.

src/yuv_convert.cc
skal65535: verified

src/yuv_convert.cc
```diff
@@ -545,13 +549,13 @@ static void ConvertWRGBToYUV(const fixed_y_t* best_y,
...
                              uint8_t* y_plane,
                              uint8_t* u_plane, uint8_t* v_plane) {
-  const int w = (width + 1) & ~1;
-  const int h = (height + 1) & ~1;
-  const int uv_w = w >> 1;
-  const int uv_h = h >> 1;
-  for (int j = 0; j < height; ++j) {
-    const int off = (j >> 1) * 3 * uv_w;
-    for (int i = 0; i < width; ++i) {
+  const size_t w = ((size_t)width + 1) & ~1ULL;
```
jzern, line 552: `... ~size_t{1}` might be more type correct.

skal65535: it's kind of hard to read... (and type promotion applies, no?)

PR #164, opened 2026-08-23
Add AVX2 support and use it for StoreHisto, QuantizeBlock, and SjpegRiskiness

skal65535:
* new flags and options: HAVE_AVX2 (Makefile) / SJPEG_HAVE_AVX2
  (CMake/Android.mk), off by default
* Picked at runtime via SupportsAVX2(), so a binary built with it still runs on pre-Haswell CPUs.
* new files: src/histogram_avx2.cc, src/quantize_avx2.cc, src/riskiness_avx2.cc
- StoreHisto: just 256-bit widening of StoreHistoSSE2. Measured -10.9% to -18.4% faster
- QuantizeBlock: Measured -3.9% faster. scalar zig-zag/run-level tail is still the slow part.
- Includes a commented-out SJPEG_USE_PEXT toggle for BMI2's PEXT
  (off by default — it's a risky instruction, esp. on AMD Zen1/Zen+)
- SjpegRiskiness: AVX2 gather for the score-table lookups, behind its own
  SJPEG_USE_AVX2_RISKINESS toggle (off by default).
  Measured -44% to -60% function speedup, -2.5% to -3.3% end to end.

- Overall (60 photos test, -q 0/75/95, AUTO mode, same x86 box): flat at q=0, -1.5% to -9% at q=75/95.
- Bit-exact against the existing SSE2/NEON/C paths in every test run
- Tested with parallel encoding too, for frequency scaling. Nothing conclusive to report.

Still TODO:
- Fdct-AVX2: need a totally new 256-bit shuffle/transpose

### DIFF
--- a/src/histogram_avx2.cc
+++ b/src/histogram_avx2.cc
@@ -1,0 +1,54 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//  AVX2 variant of StoreHisto(), compiled with -mavx2 only for this file
+//  (see Makefile's src/%_avx2.o rule) so the rest of the library doesn't
+//  need to be built for an AVX2 target. 256-bit widening of StoreHistoSSE2
+//  in histogram.cc: same per-lane abs/shift/min, no cross-lane shuffle, so
+//  it doubles lane count 8 -> 16 with no algorithmic change. The trailing
+//  per-coefficient histogram increment stays scalar either way (data-
+//  dependent scatter, not vectorizable with AVX2).
+//
+// Author: Skal (pascal.massimino@gmail.com)
+
+#define SJPEG_NEED_ASM_HEADERS
+#include "sjpegi.h"
+
+#if defined(SJPEG_USE_AVX2)
+
+namespace sjpeg {
+
+void StoreHistoAVX2(const int16_t in[64], Histo* const histos,
+                    int nb_blocks) {
+  const __m256i kMaxHisto = _mm256_set1_epi16(MAX_HISTO_DCT_COEFF);
+  for (int n = 0; n < nb_blocks; ++n, in += 64) {
+    uint16_t tmp[64];
+    for (int i = 0; i < 64; i += 16) {
+      const __m256i A =
+          _mm256_loadu_si256(reinterpret_cast<const __m256i*>(in + i));
+      const __m256i C = _mm256_abs_epi16(A);
+      const __m256i D = _mm256_srli_epi16(C, HSHIFT);
+      const __m256i E = _mm256_min_epi16(D, kMaxHisto);
+      _mm256_storeu_si256(reinterpret_cast<__m256i*>(tmp + i), E);
+    }
+    for (int j = 0; j < 64; ++j) {
+      const int k = tmp[j];
+      ++histos->counts_[j][k];
+    }
+  }
+}
+
+}  // namespace sjpeg
+
+#endif  // SJPEG_USE_AVX2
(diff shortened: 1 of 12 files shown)


hasinoff:
When I was looking a bit into AVX2, one quick win I found was to bump ALIGN_CST to 31.
Does that help here?
(I think it would worthwhile to fold that change in regardless)

skal65535:
> When I was looking a bit into AVX2, one quick win I found was to bump ALIGN_CST to 31. Does that help here? (I think it would worthwhile to fold that change in regardless)

So i tried to change ALIGN_CST to 31 or 63, and it didn't change the timing. Even using aligned-loads in the SIMD code didn't make a difference (this patch uses unaligned-loads).

hasinoff:
> > When I was looking a bit into AVX2, one quick win I found was to bump ALIGN_CST to 31. Does that help here? (I think it would worthwhile to fold that change in regardless)
> 
> So i tried to change ALIGN_CST to 31 or 63, and it didn't change the timing. Even using aligned-loads in the SIMD code didn't make a difference (this patch uses unaligned-loads).

Fair enough, maybe this only helps with e.g. 256-bit shuffle masks for color conversion with alignas(32).

src/histogram_avx2.cc
```diff
@@ -0,0 +1,54 @@
...
+#include "sjpegi.h"
+
+#if defined(SJPEG_USE_AVX2)
+
+namespace sjpeg {
+
+void StoreHistoAVX2(const int16_t in[64], Histo* const histos,
+                    int nb_blocks) {
+  const __m256i kMaxHisto = _mm256_set1_epi16(MAX_HISTO_DCT_COEFF);
+  for (int n = 0; n < nb_blocks; ++n, in += 64) {
```
jzern:
```
assert(nb_blocks > 0);
do {
...
} while
```

should work.

src/histogram_avx2.cc
```diff
@@ -0,0 +1,54 @@
...
+    uint16_t tmp[64];
+    for (int i = 0; i < 64; i += 16) {
+      const __m256i A =
+          _mm256_loadu_si256(reinterpret_cast<const __m256i*>(in + i));
```
jzern: Not necessary, but `LOAD_32` / `STORE_32` could be added to match the sse2.

src/quantize_avx2.cc
```diff
@@ -0,0 +1,108 @@
...
+    const __m256i D = _mm256_adds_epi16(C, m_bias);            // v' = v+bias
+    const __m256i E = _mm256_mulhi_epu16(D, m_mult);           // (v'*iq)>>16
+    const __m256i F = _mm256_srli_epi16(E, AC_BITS);           // QUANTIZE(...)
+    const __m256i G = _mm256_xor_si256(F, B);                  // v ^ mask
+    _mm256_storeu_si256(reinterpret_cast<__m256i*>(tmp + i), F);
+    _mm256_storeu_si256(reinterpret_cast<__m256i*>(masked + i), G);
+    const __m256i cmp = _mm256_cmpgt_epi16(F, zero);
+#if defined(SJPEG_USE_PEXT)
+    const uint32_t m32 = static_cast<uint32_t>(_mm256_movemask_epi8(cmp));
+    const uint32_t m16 = static_cast<uint32_t>(_pext_u32(m32, 0x55555555u));
```
jzern, line 74: This would require another runtime check, requires BMI.

jzern (approved):
`StoreHisto` and `QuantizeBlock` are close to what I had locally. I didn't review the riskiness part yet.

src/quantize_avx2.cc
skal65535: yes, that's why i leave it as an uncommented #define. It's not really making things super faster, so it's more for documentation than practical use. Could be removed.

src/histogram_avx2.cc
skal65535: done

src/histogram_avx2.cc
skal65535: done

src/histogram_avx2.cc
```diff
@@ -29,26 +29,35 @@
 
 namespace sjpeg {
 
+#define LOAD_32(src) _mm256_loadu_si256(reinterpret_cast<const __m256i*>(src))
+#define STORE_32(V, dst) \
+  _mm256_storeu_si256(reinterpret_cast<__m256i*>(dst), (V))
+
 void StoreHistoAVX2(const int16_t in[64], Histo* const histos,
                     int nb_blocks) {
+  assert(nb_blocks > 0);
```
jzern, line 40: add `<assert.h>`.

jzern (approved):

src/histogram_avx2.cc
skal65535: done

PR #166, opened 2026-08-24
Sharp YUV: early out at first iteration if already converged

skal65535:
Do a convergence test ("diff_y_sum < diff_y_threshold") right after
the 1rst iteration, for an early out.
The 'things are getting worse' test stays active after that, no change.

test: -32% on ApplySharpYUVConversion itself, isolated, over 300 test photos.

measured -8.7% speed-up end-to-end encode with default params.

it's not a bit-exact change, but measured difference is insignificant (<0.05dB)

### DIFF
--- a/src/yuv_convert.cc
+++ b/src/yuv_convert.cc
@@ -674,10 +674,8 @@ static bool PreprocessARGB(const uint8_t* const rgb, int width, int height,
                       (int)(3 * uv_w));
     }
     // test exit condition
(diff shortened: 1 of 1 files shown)


jzern (approved):
cc: @maryla-uc

skal65535:
> cc: @maryla-uc

https://chromium-review.googlesource.com/c/webm/libwebp/+/8316388

PR #169, opened 2026-08-24
sjpegi.h: remove unused NULL check/define

jzern:
`NULL` is not used in the code base since:
 927069e Replace NULL and 0 with nullptr in C++ code (#137)

The check itself was likely never needed unless there was a missed C
system include.


skal65535:
LGTM

PR #171, opened 2026-08-24
sync README.md w/README; delete README

jzern:
- **base**
- **sync README.md w/README; delete README**


skal65535 (approved):
READMEs are soooo 2010

PR #173, opened 2026-08-26
Progressive coding: memory-safety and clarity fixes

skal65535:
- AllocateProgPlanes(): guard run_levels sizing against 32-bit overflow
- EncodeProgressive(): free prog_planes_ on the two DC-scan error
  paths that skipped it
- EncodeProgressive(): free in_blocks_ right after the quantize pass
- EncodeProgressive(): free each component's storage right after its
  own AC scan
- SinglePassEncode(): check EncodeProgressive()'s return value instead
  of relying on the ok_ side effect
- sjpeg.h: document that -prog is dropped under multi-pass search
- SinglePassEncode(): note the progressive dispatch's ordering
  constraint
- CheckProgBuffers(): document why it skips all_run_levels_ growth
- extract the shared have_coeffs_/TransformMCU() guard into
  MaybeTransformMCU()
- EncodeProgressive(): fold the DC-stats pass into the quantize pass
- drop the redundant progressive_ field (prog_luma_split_ < 64 says
  the same thing)
- add FreePtr(): Free()+null a pointer in one call, applied at
  existing call sites

### DIFF
--- a/src/sjpegi.h
+++ b/src/sjpegi.h
@@ -414,6 +414,10 @@ struct Encoder {
   // and progressive quantize loops.
   void TransformMCU(int mb_x, int mb_y, int16_t* const out);
 
+  // TransformMCU() unless coefficients are already cached. Shared guard for
+  // SinglePassScan(), SinglePassScanOptimized(), EncodeProgressive().
+  void MaybeTransformMCU(int mb_x, int mb_y, int16_t** const in);
+
   // collect transformed coeffs (unquantized) only
   void CollectCoeffs();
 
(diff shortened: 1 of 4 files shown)


src/sjpegi.h
```diff
@@ -519,6 +523,11 @@ struct Encoder {
   template<class T> void Free(T* const ptr) {
     memory_hook_->Free(reinterpret_cast<void*>(ptr));
   }
+  // Free(*ptr) followed by '*ptr = nullptr', as a single safe step.
+  template<class T> void FreePtr(T** const ptr) {
```
jzern, line 527: A later improvement might be to just switch `Alloc` to return `std::unique_ptr`.

jzern (approved):

src/sjpegi.h
skal65535: agreed

skal65535: @jzern actually, i turns out some detail outweight the benefits:
 * we have memory_hook_, so we'd need special deleter instead of std::default_delete  => more complex signature and code-dragging
 * it would mostly trade a FreePtr() call against a x.reset() call on unique_ptr -> no much RAII and clarity in fact.

jzern: Good point, I was thinking the memory_hook_ would cause some complications and you're right, given how it's used you'd replace one call for another. At most you'd only satisfy some code analysis / clang-tidy.

PR #177, opened 2026-08-28
Add adaptive-bias quantization

skal65535:
'-adapt_bias' modulates the AC quantization bias per-block.

* adds Encoder::AdaptiveBiasQuantizeBlock, dispatched with GetActiveQuantizeBlockFunc()
* zero cost to the non-adaptive paths.

JPEG only spends bits on coeffs' num_bits coding, mantissa isn't Huffman-coded.
=> adaptive bias only nudges coefficients at the threshold of power-of-two.
And by an amount proportional to a per-block flat/normal/busy activity
estimation.


skal65535:
should probably wait for the multi-thread code to be merged first, i'll be easier to rebase

examples/vjpeg.cc
```diff
@@ -302,9 +324,11 @@ static void ComputeErrorMap() {
 }
 
 namespace sjpeg {
-// undocumented function
+// undocumented functions
```
jzern: This comment is a little weird unless it was working around some lint warning.

examples/vjpeg.cc
```diff
@@ -339,6 +363,72 @@ static void ComputeRiskinessMap() {
   }
 }
 
+static void ComputeActivityMap() {
+  if (kParams.show != 7) {
+    const int width = kParams.width;
+    const int height = kParams.height;
+    const int stride = 3 * width;
```
jzern, line 374: Hopefully no huge image.

src/sjpegi.h
```diff
@@ -299,6 +301,20 @@ extern MemoryManager* GetDefaultMemoryManager();
 // a division, AC_BITS for the extra precision left over by the fdct's scaling.
 enum { FP_BITS = 16, AC_BITS = 4 };
 
+// Block activity classification for perceptual adaptive-bias quantization.
+enum BlockActivityTier { kFlatBlock = -1, kNormalBlock = 0, kBusyBlock = 1 };
```
jzern: If `BlockActivityScore` returns the enum then I don't think the values are used directly, so this could be an `enum class`.

src/jpeg_tools.cc
```diff
@@ -469,4 +469,31 @@ double BlockRiskinessScore(const uint8_t* rgb, int stride,
...
+  for (int i = 1; i < 64; ++i) {
+    activity += std::abs(in[i]);
+  }
+  if (score != nullptr) *score = activity;
+  return ClassifyActivity(activity);
+}
+
+// This function returns the block activity classification:
+// -1 = flat, 0 = normal, 1 = busy.
+int BlockActivityScore(const uint8_t* rgb, int stride,
```
jzern: This can return `BlockActivityTier` too.

tests/unit_test.cc
```diff
@@ -408,8 +408,8 @@ SJPEG_TEST(NegativeStrides) {
   match();
 
   SJPEG_CHECK(sjpeg::EncodeNV12(Flip(Y, W, H).data(), W,
-                                Flip(UV, uv_stride, uv_h).data(), uv_stride,
-                                W, H, p, sjpeg::MakeByteSink(&a).get()));
+                                 Flip(UV, uv_stride, uv_h).data(), uv_stride,
```
jzern: bad indent

jzern (approved):

examples/vjpeg.cc
skal65535: removed

examples/vjpeg.cc
skal65535: int -> size_t

src/sjpegi.h
skal65535: indeed, done

src/jpeg_tools.cc
skal65535: done

tests/unit_test.cc
skal65535: fixed

PR #179, opened 2026-08-29
~2.6x faster progressive riskiness scoring

skal65535:
Replaces #167's fixed-margin/checkpoint scheme with a windowed-stddev
stopping criterion.

* `MovingWindow`: 32-sample running mean/stddev sizes the exit margin
  per metric. Checked after every band.
* Three values are tracked as the scan progresses, and we only stop
  once each one has stopped moving around: how much of the image has
  enough signal to trust the score at all, whether the image is gray
  enough to skip chroma entirely, and the sharpness score itself
  against its two thresholds. Gray gets checked first since it
  overrides everything else.
* `SjpegRiskiness()` name/signature unchanged; early exit is the new
  default.
* No public always-exact entry point added -- `SjpegRiskinessFullForEval()`
  is internal-only, used solely to produce the before/after eval numbers
  below.

Eval: 2422 images (1863 real photos + 259 synthetic), **0 flips**, max
score diff 4.82, up to 2.6x speedup.

### DIFF
--- a/src/jpeg_tools.cc
+++ b/src/jpeg_tools.cc
@@ -179,14 +179,54 @@ static const double kThreshYU420 = 40.0;
 static const double kThreshSharpYU420 = 70.0;
 
 // Progressive refinement: score rows in bands spread in bit-reversed dispersed
-// order. Check score at pre-defined increasingly sparse check-points.
-// Equivalent to a sequential pass if there's no early-out (full coverage).
+// order. After *every* band, check whether the running estimate has settled
+// far enough from the decision cutoffs to stop early. Equivalent to a
+// sequential pass if there's no early-out (full coverage).
 static const int kBandHeight = 8;
-static const double kConfidenceMargin = 4.0;
-static const int kNumCheckpoints = 4;
-static const double kCheckpointFractions[kNumCheckpoints + 1] = {
-  0.05, 0.15, 0.30, 0.50, /*sentinel*/1.0
+
+// Running std-dev over the last kWindowSize pushed values, used to size how
+// far from a cutoff a metric must sit before it's trusted as "settled".
+class MovingWindow {
+ public:
+  MovingWindow() { memset(values_, 0, sizeof(values_)); }
+  // Pushes 'value' and returns the resulting margin: HUGE_VAL until the
+  // window is full, else z_score * std-dev (floored at margin_floor).
+  double Push(double value, double z_score, double margin_floor) {
+    if (count_ == kWindowSize) {
+      const double old = values_[next_];
+      sum_ -= old;
+      sumsq_ -= old * old;
+    } else {
+      ++count_;
+    }
+    values_[next_] = value;
+    sum_ += value;
+    sumsq_ += value * value;
+    next_ = (next_ + 1) % kWindowSize;
+
+    if (count_ < kWindowSize) return HUGE_VAL;
+    const double mean = sum_ / count_;
+    const double variance = sumsq_ / count_ - mean * mean;
+    const double margin = z_score * sqrt((variance > 0.) ? variance : 0.);
+    return (margin > margin_floor) ? margin : margin_floor;
+  }
+
+ private:
+  static const int kWindowSize = 32;
+  double values_[kWindowSize];
+  double sum_ = 0., sumsq_ = 0.;
+  int count_ = 0, next_ = 0;
 };
+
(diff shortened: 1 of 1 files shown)


src/jpeg_tools.cc
```diff
@@ -179,15 +179,55 @@ static const double kThreshYU420 = 40.0;
...
-static const double kConfidenceMargin = 4.0;
-static const int kNumCheckpoints = 4;
-static const double kCheckpointFractions[kNumCheckpoints + 1] = {
-  0.05, 0.15, 0.30, 0.50, /*sentinel*/1.0
+
+// Running std-dev over the last kWindowSize pushed values, used to size how
+// far from a cutoff a metric must sit before it's trusted as "settled".
+class MovingWindow {
+ public:
+  MovingWindow() { memset(values_, 0, sizeof(values_)); }
```
jzern: An alternative could be:

```
 public:
  MovingWindow() = default;
...
 private:
  ...
  double values_[kWindowSize] = {};
```

src/jpeg_tools.cc
```diff
@@ -179,15 +179,55 @@ static const double kThreshYU420 = 40.0;
...
+
+    if (count_ < kWindowSize) return HUGE_VAL;
+    const double mean = sum_ / count_;
+    const double variance = sumsq_ / count_ - mean * mean;
+    const double margin = z_score * sqrt((variance > 0.) ? variance : 0.);
+    return (margin > margin_floor) ? margin : margin_floor;
+  }
+
+ private:
+  static const int kWindowSize = 32;
```
jzern: could be `constexpr`.

src/jpeg_tools.cc
```diff
@@ -210,8 +250,12 @@ extern int RiskinessScoreRowAVX2(const uint16_t* row1, const uint16_t* row2,
 }  // namespace sjpeg
 #endif
 
-SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
-                            int width, int height, int stride, float* risk) {
+// if 'full_scan' is true, the progressive early-exit is disabled and every
+// band gets scored -- used only to produce a before/after reference for
+// evaluation, not exposed in the public API.
+static SjpegYUVMode RiskinessImpl(const uint8_t* rgb,
+                                   int width, int height, int stride,
```
jzern: indent is off

src/jpeg_tools.cc
```diff
@@ -316,23 +363,46 @@ SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
...
+          fabs(candidate_risk - kThreshSharpYU420) >= score_margin) {
         if (risk != nullptr) *risk = candidate_risk;
         return candidate_mode;
       }
-      ++next_checkpoint;
-      target = (int)ceil(kCheckpointFractions[next_checkpoint] * num_bands);
     }
   }
-  return Finalize(risk);
+  return Finalize(risk, nullptr, nullptr);
```
jzern: Could mark the `nullptr` parameters: `..., /*gray_count_out=*/nullptr, /*frac_out=*/nullptr);`

src/jpeg_tools.cc
```diff
@@ -316,23 +363,46 @@ SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
...
+
+SjpegYUVMode SjpegRiskiness(const uint8_t* rgb,
+                            int width, int height, int stride, float* risk) {
+  return RiskinessImpl(rgb, width, height, stride, risk, /*full_scan=*/false);
+}
+
+// internal-only entry point for before/after evaluation, not declared in the
+// public header.
+SjpegYUVMode SjpegRiskinessFullForEval(const uint8_t* rgb,
+                                        int width, int height, int stride,
```
jzern: indent is off

jzern (approved):

src/jpeg_tools.cc
skal65535: simpler -> done

src/jpeg_tools.cc
skal65535: fixed

src/jpeg_tools.cc
skal65535: done

src/jpeg_tools.cc
skal65535: fixed

src/jpeg_tools.cc
skal65535: done

PR #180, opened 2026-08-29
namespace checks

jzern:
- **unit_test.cc: namespace CHECK**
- **unit_test.cc: namespace TEST**


jzern:
Up to you if you want to handle it this way. Saves a bunch of noise like: "you should be using CHECK_EQ (and you forgot  the include/dependency", or "missing googletest include", or "looks like you're bad at this, would you like clippy^w\<model\> to fix it for you?"

skal65535 (approved):
fool-proof indeed </clippy>

PR #181, opened 2026-08-31
Add gather-based AVX2 variant of SharpYUV gamma-table lookups

skal65535:
* UpdateW/UpdateChroma AVX2 version uses vgather for the GammaToLinear/LinearToGamma lookups
* Bit-exact same as the C reference
* ~1.13x faster on ApplySharpYUVConversion alone
* ~1.09x end-to-end with default SjpegCompress params
   (SJPEG_YUV_AUTO only routes some images through SharpYUV)
* new gates: SJPEG_HAVE_AVX2 + SJPEG_USE_AVX2_YUV_GATHER (on by default)
* dynamic runtime dispatch with SupportsAVX2().

### DIFF
--- a/src/yuv_convert.cc
+++ b/src/yuv_convert.cc
@@ -38,8 +38,6 @@ namespace sjpeg {
 #define SFIX 2                // fixed-point precision of RGB and Y/W
 #define SHALF (1 << SFIX >> 1)
 #define MAX_Y_T ((256 << SFIX) - 1)
-typedef int16_t fixed_t;      // signed type with extra SFIX precision for UV
-typedef uint16_t fixed_y_t;   // unsigned type with extra SFIX precision for W
 
 static fixed_y_t clip_y(int y) {
   return (!(y & ~MAX_Y_T)) ? (fixed_y_t)y : (y < 0) ? 0 : MAX_Y_T;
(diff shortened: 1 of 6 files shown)


src/yuv_convert.cc
```diff
@@ -430,18 +449,24 @@ static void InitFunctionPointers() {
       kSharpUpdateRGB = SharpUpdateRGB_NEON;
       kSharpFilterRow = SharpFilterRow_NEON;
     }
+#endif
+#if defined(SJPEG_HAVE_AVX2) && defined(SJPEG_USE_AVX2_YUV_GATHER)
+    if (sjpeg::SupportsAVX2()) {
+      kUpdateW = UpdateWAVX2;
```
jzern: Should be `_AVX2` to match the other functions.

src/yuv_convert_avx2.cc
```diff
@@ -0,0 +1,165 @@
...
+      reinterpret_cast<const int*>(kLinearToGammaTab), tab_pos, 4);
+  const __m256i v1 = _mm256_i32gather_epi32(
+      reinterpret_cast<const int*>(kLinearToGammaTab),
+      _mm256_add_epi32(tab_pos, _mm256_set1_epi32(1)), 4);
+  const __m256i v2 = _mm256_mullo_epi32(_mm256_sub_epi32(v1, v0), x);
+  return _mm256_add_epi32(v0, _mm256_srli_epi32(v2, kGammaToLinearBits));
+}
+
+static inline __m256i RGBToGray8(__m256i r, __m256i g, __m256i b) {
+  const __m256i round = _mm256_set1_epi32(1 << 16 >> 1);
```
jzern: Could just be `1 << 15` for simplicity.

jzern (approved):

src/yuv_convert.cc
skal65535: ah yes, done.

src/yuv_convert_avx2.cc
skal65535: done

Issue #163, opened 2026-08-23
add SSSE3, SSE4.1, AVX, AVX2 support


skal65535:
Not sure we need all the relatively old variants, but at least SSE4.1 and AVX2 would be good anchor points

skal65535:
PR #164

skal65535:
For the record, here are the top functions:

### Top 10, no SIMD (pure C)

| # | % self | Function |
|---|--------|----------|
| 1 | 21.1% | `ApplySharpYUVConversion` |
| 2 | 16.6% | `QuantizeBlock` |
| 3 | 10.7% | `StoreHisto` |
| 4 | 10.6% | `SjpegRiskiness` |
| 5 | 8.6% | `RowToIndexC` |
| 6 | 8.4% | `FdctC` |
| 7 | 6.2% | `Get16x8Block_C` |
| 8 | 4.8% | `Encoder::CodeBlock` |
| 9 | 3.6% | `SharpFilterRow_C` |
| 10 | 1.9% | `Encoder::AddEntropyStats` |

### Top 10, with SIMD (NEON, current default)

| # | % self | Function |
|---|--------|----------|
| 1 | 29.1% | `ApplySharpYUVConversion` |
| 2 | 14.6% | `SjpegRiskiness` |
| 3 | 11.7% | `StoreHistoNEON` |
| 4 | 9.2% | `QuantizeBlockNEON` |
| 5 | 7.8% | `RowToIndexNEON` |
| 6 | 6.5% | `Encoder::CodeBlock` |
| 7 | 6.1% | `FdctNEON` |
| 8 | 2.5% | `Get16x8_NEON` |
| 9 | 2.5% | `SharpFilterRow_NEON` |
| 10 | 1.7% | `Encoder::AddEntropyStats` |

skal65535:
good chunk of optims added. Still need to decide if we drop SSSE3 in favor of a full SSE4.1 variants

jzern:
> good chunk of optims added. Still need to decide if we drop SSSE3 in favor of a full SSE4.1 variants

I think that's a good way to go considering where this library is (and may be) used.

jzern:
Link errors with `clang-cl` at `8d73443`.

```
> cmake .. -DSJPEG_HAVE_AVX2=1 -T ClangCl
> cmake --build .
...
lld-link : error : undefined symbol: __cpu_indicator_init [sjpeg-bin.vcxproj]
  >>> referenced by src\enc.cc:206
[...]

PR #185, opened 2026-09-02
enc.cc,msvc: add clang_rt.builtins for __builtin_cpu_*

jzern:
Fixes link errors related to `__builtin_cpu_init` and
`__builtin_cpu_supports` when using `clang-cl.exe` or `clang.exe` to
target the Windows ABI.

Issue #163

### DIFF
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -203,6 +203,13 @@ bool SupportsNEON() {
 bool SupportsAVX2() {
   if (ForceSlowCImplementation) return false;
 #if defined(SJPEG_HAVE_AVX2) && (defined(__i386__) || defined(__x86_64__))
+#if defined(__clang__) && defined(_MSC_VER)
+#if defined(__x86_64__)
+#pragma comment(lib, "clang_rt.builtins-x86_64.lib")
+#else
+#pragma comment(lib, "clang_rt.builtins-i386.lib")
+#endif
+#endif
   __builtin_cpu_init();
   return __builtin_cpu_supports("avx2") != 0;
 #else


jzern:
Detection in CMake looked something like:

```
+  if(CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC" AND
+     CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    # Add clang_rt.builtins-*.lib for __builtin_cpu_*().
+    if(CMAKE_CXX_COMPILER_ARCHITECTURE_ID STREQUAL "X86")
+      target_link_options(sjpeg PUBLIC
+                          "/DEFAULTLIB:clang_rt.builtins-i386.lib")
+    elseif(CMAKE_CXX_COMPILER_ARCHITECTURE_ID STREQUAL "x64")
+      target_link_options(sjpeg PUBLIC
+                          "/DEFAULTLIB:clang_rt.builtins-x86_64.lib")
+    endif()
+  endif()
```

But this only covered `clang-cl`, not `clang`.

src/enc.cc
```diff
@@ -203,6 +203,13 @@ bool SupportsNEON() {
 bool SupportsAVX2() {
   if (ForceSlowCImplementation) return false;
 #if defined(SJPEG_HAVE_AVX2) && (defined(__i386__) || defined(__x86_64__))
+#if defined(__clang__) && defined(_MSC_VER)
+#if defined(__x86_64__)
+#pragma comment(lib, "clang_rt.builtins-x86_64.lib")
+#else
+#pragma comment(lib, "clang_rt.builtins-i386.lib")
+#endif
+#endif
```
hasinoff, line 212: Ideally non-clang Windows would return false - this will fail to compile on MSVC, I think.

jzern: MSVC (`cl.exe`) will not have `__i386__` or `__x86_64__` defined.

jzern: > MSVC (`cl.exe`) will not have `__i386__` or `__x86_64__` defined.

It fails to link due to the deficient SSE2 check, mentioned in https://github.com/webmproject/sjpeg/issues/163#issuecomment-5504774267.

jzern: This works in the 3 configurations I tested with Visual Studio 2026:
* `-G Visual Studio 18 2026 -T ClangCl` (`clang-cl.exe`)
* `-G Visual Studio 18 2026` (`cl.exe`, hits the `#else`)
* `-G Ninja Multi-Config` (`clang.exe`)

So, later some work can be done to add `cl.exe` support using the `__cpuid(ex)` and `_xgetbv` intrinsics like we do in libwebp.

skal65535: sgtm

skal65535 (approved):

PR #187, opened 2026-09-12
Optimize ARM NEON histogramming and Trellis dynamic programming search

hasinoff:
# Pull Request: Optimize ARM NEON Histogram and Trellis Quantization

## Title
`Optimize ARM NEON histogramming and Trellis dynamic programming search`

## Summary
This PR introduces two surgical optimizations targeting hot inner loops in `sjpeg`:

### 1. Direct Register Lane Extraction in `StoreHistoNEON` (`src/histogram.cc`)
* **Problem:** `StoreHistoNEON` was writing 128-bit vector results into a 64-element local stack array (`tmp[64]`) via `vst1q_u16`, and immediately reading back scalar values `tmp[j]` in a subsequent scalar loop to increment histogram buckets.
* **Optimization:** Replaced the stack memory round-trip with direct register extraction using `vgetq_lane_u16(E, 0..7)`.
* **Benefit:** Completely eliminates the 64-byte stack allocation and avoids 8 sequential Store-to-Load Forwarding (STLF) stalls per 8x8 block on out-of-order execution pipelines.

### 2. Invariant Hoisting in `SearchBestPrev` (`src/quantize.cc`)
* **Problem:** In Method 7 (Trellis quantization), the backward dynamic programming search `SearchBestPrev` repeatedly re-evaluated `(codes[0xf0] & 0xff)` inside the inner `for (const TrellisNode* cur = node - 1; cur >= nodes0; --cur)` loop.
* **Optimization:** Hoisted `const uint32_t zrl_len = codes[0xf0] & 0xff;` outside the loop.
* **Benefit:** Reduces loop register pressure and redundant indexed lookups across millions of node evaluations.

---

### 1. Kernel Microbenchmarks
| Kernel | Platform | Baseline | Optimized | Speedup |
| :--- | :--- | :--- | :--- | :--- |
| **`StoreHistoNEON`** | ARM Neoverse V2 | 26.69 Mblocks/s | **32.45 Mblocks/s** | **+21.6%** |
| **`SearchBestPrev`** (Trellis) | x86-64 (Milan) | 52.1 ns/block | **42.2 ns/block** | **+19.0%** |
| **`SearchBestPrev`** (Trellis) | ARM Neoverse V2 | 48.4 ns/block | **39.4 ns/block** | **+18.6%** |

### 2. End-to-End Encoding Speedups
| Encoding Mode | Architecture | E2E Speedup |
| :--- | :--- | :--- |
| **Method 1 (Optimized Huffman)** | ARM64 NEON | **+0.8% to +1.5%** |
| **Method 7 (Trellis Quantization)** | x86-64 | **+4.5% to +6.8%** |
| **Method 7 (Trellis Quantization)** | ARM64 NEON | **+5.2% to +7.4%** |

---

## Correctness & Bit-Exact Parity
* Bit-exact output verified across the full test suite (`unit_test`, `test_api`, `test_param`, `test_io`, etc.) with 0 differences in output JPEG streams.

### DIFF
--- a/src/histogram.cc
+++ b/src/histogram.cc
@@ -80,18 +80,20 @@ void StoreHistoSSE2(const int16_t in[64], Histo* const histos, int nb_blocks) {
 void StoreHistoNEON(const int16_t in[64], Histo* const histos, int nb_blocks) {
   const uint16x8_t kMaxHisto = vdupq_n_u16(MAX_HISTO_DCT_COEFF);
   for (int n = 0; n < nb_blocks; ++n, in += 64) {
-    uint16_t tmp[64];
     for (int i = 0; i < 64; i += 8) {
       const int16x8_t A = vld1q_s16(in + i);
       const int16x8_t B = vabsq_s16(A);               // abs(in)
       const uint16x8_t C = vreinterpretq_u16_s16(B);  // signed->unsigned
       const uint16x8_t D = vshrq_n_u16(C, HSHIFT);    // >>= HSHIFT
       const uint16x8_t E = vminq_u16(D, kMaxHisto);   // min(.,kMaxHisto)
-      vst1q_u16(tmp + i, E);
-    }
-    for (int j = 0; j < 64; ++j) {
-      const int k = tmp[j];
-      ++histos->counts_[j][k];
+      ++histos->counts_[i + 0][vgetq_lane_u16(E, 0)];
+      ++histos->counts_[i + 1][vgetq_lane_u16(E, 1)];
+      ++histos->counts_[i + 2][vgetq_lane_u16(E, 2)];
+      ++histos->counts_[i + 3][vgetq_lane_u16(E, 3)];
+      ++histos->counts_[i + 4][vgetq_lane_u16(E, 4)];
+      ++histos->counts_[i + 5][vgetq_lane_u16(E, 5)];
+      ++histos->counts_[i + 6][vgetq_lane_u16(E, 6)];
+      ++histos->counts_[i + 7][vgetq_lane_u16(E, 7)];
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/histogram.cc
```diff
@@ -80,18 +80,20 @@ void StoreHistoSSE2(const int16_t in[64], Histo* const histos, int nb_blocks) {
...
       const int16x8_t B = vabsq_s16(A);               // abs(in)
       const uint16x8_t C = vreinterpretq_u16_s16(B);  // signed->unsigned
       const uint16x8_t D = vshrq_n_u16(C, HSHIFT);    // >>= HSHIFT
       const uint16x8_t E = vminq_u16(D, kMaxHisto);   // min(.,kMaxHisto)
-      vst1q_u16(tmp + i, E);
-    }
-    for (int j = 0; j < 64; ++j) {
-      const int k = tmp[j];
-      ++histos->counts_[j][k];
+      ++histos->counts_[i + 0][vgetq_lane_u16(E, 0)];
```
skal65535: wonder if extracting 'histos->counts_' pointer outside of the loop as a const immutable pointer would help...

skal65535: (with a 'restrict' even?)

skal65535 (approved):
(left some comment)

src/histogram.cc
hasinoff: That helped a bit.
- Kernel Microbenchmark (ARM Neoverse V2): Throughput improves by +5.9% (from 32.45 Mblocks/s to 34.36 Mblocks/s), representing a cumulative +28.7% speedup over the original tmp[64] stack-spill baseline.

`__restrict__` has zero effect because of strict aliasing rules - `in` and `counts` are different scalar types, so the compiler already knows they cannot alias in memory.

skal65535 (approved):

PR #188, opened 2026-09-12
NEON fdct: avoid 16-bit intermediate overflow on bright blocks

skal65535:
Factoring out shared products in the horizontal even part using 16-bit
butterflies causes intermediate additions to overflow on bright or unshifted
blocks in [0, 255] (a0 + a3 can exceed 32767 and wrap around modulo 65536),
inverting the DC coefficient and triggering an assert in entropy coding.

Instead of 16-bit addition, accumulate ps03 and ps12 directly in 32-bit
using vmull_s16 + vmlal_s16 (MULT_DCL_32 + MULT_ADD_32). Keep the 16-bit
subtractions for d03 and d12 since DC cancelation keeps differences strictly
within int16_t bounds.

Performance (Apple Silicon M-series, 5M blocks):
- Baseline (before optimization): 26.5 ns / block
- 16-bit butterfly (buggy):        23.6 ns / block
- 32-bit MAC (fixed):              23.7 ns / block (+0.13 ns delta, retains 95% of gain)

Add a regression test in unit_test.cc exercising unshifted bright blocks.

The problem is seen with U=128 or V=128, that can happen with the current RGB->YUV rounding constant HALF.
Alternative fix:
      a) if we use HALF-1 rounding, U and V are 127 at most, and the saturation issue disappears
      b) clamp U/V input to 127 max (forcing s8 input, s16 output), but the  number of call sites to update and the fact that output bitstream will be different, makes it impractical.


skal65535:
Note: the real issue is that, with the current coeffs and rounding for RGB->YUV conversion, the input for U/V is [-127, 128] range. For instance, for pure blue or pure red input.
We can fix that by using `HALF-1` instead of `HALF` as rounding constants, making sure the input to fdct is [-128,127]. This little change makes the overflow disappear and we don't need to use the fix in this PR.

But! It also means a lot of golden and reference bitstream will differ slightly (measured impact is < 0.002% in size or dB in my tests). And that could be a problem, to update all tests and hardcoded bitstreams.

skal65535:
Another possibility i haven't fully explored yet is forcing the fdct input to be s8, with range clamped to [-127,127], to avoid the problematic +128 upper input value.

This change, however, has a quite large blast radius: we'd need a different in/out buffer (s8->int16), add extra s8->s16 uplift instructions, etc.
The net cost is much larger than the two extra instructions in the 32b butterfly. Not worth it.

src/fdct.cc
```diff
@@ -506,11 +506,15 @@ static void Dct_NEON(int16_t* in) {
...
-  BUTTERFLY(d03, s03, a0, a3);
-  BUTTERFLY(d12, s12, a1, a2);
-
-  MULT_DCL_32(ps03_lo, ps03_hi, s03, kC4);
-  MULT_DCL_32(ps12_lo, ps12_hi, s12, kC4);
+  const int16x8_t d03 = vsubq_s16(a0, a3);
+  const int16x8_t d12 = vsubq_s16(a1, a2);
+
+  // Accumulate a0+a3 and a1+a2 in 32-bit to avoid 16-bit intermediate overflow
+  // on bright blocks (23.7 ns vs 23.6 ns / block).
```
jzern: *... (23.7 ns vs 23.6 ns / block)*: might be a little too much information since it's hardware dependent. Probably better for the commit message.

tests/unit_test.cc
```diff
@@ -25,6 +25,8 @@
 #include <thread>  // NOLINT
 #include <vector>
 
+#include <assert.h>
```
jzern: This can go with the C header group at line 18.

jzern (approved):

src/fdct.cc
skal65535: done

tests/unit_test.cc
skal65535: done

jzern (approved):

skal65535:
rebased

PR #189, opened 2026-09-13
enc: add restart marker support (DRI / RST0-RST7)

hasinoff:
EncoderParam::restart_interval_rows makes the encoder emit a DRI header and RSTn markers every N rows of MCUs, resetting the DC predictors at each interval boundary. Restart markers bound the damage a corrupted bitstream can cause and let a decoder work on intervals independently, in exchange for a couple of bytes per interval. They are off by default, so existing output is unchanged.

The interval is expressed in MCU rows rather than MCUs, matching cjpeg's -restart: sjpeg's scan loops and DC resets are row-based, and a row-aligned interval is the only kind that all of them can agree on. DRI's Ri field is 16 bits, so the row count is clamped to the coarsest whole number of rows whose MCU count still fits, following libjpeg's jcmaster.c -- restarts become more frequent than requested, never less.

Both the single-pass and the multi-pass (-size / -psnr) paths are covered. The latter writes its own headers and stores DC deltas ahead of the final scan, so it needs DRI emission and the interval DC resets wired up separately; its size estimator also accounts for the extra RSTn bytes and the padding bits that precede them, so convergence is not thrown off.

Adds a -restart option to the sjpeg tool and documents it in man/sjpeg.1, plus a unit test covering the interval-to-DRI arithmetic across resolutions, subsampling modes, Huffman and trellis quantization settings, the multi-pass search, and the 16-bit clamp.

### DIFF
--- a/man/sjpeg.1
+++ b/man/sjpeg.1
@@ -176,6 +176,15 @@ optimized for rate-distortion).
 .TP
 .B \-no_optim
 Disable Huffman code optimization (faster processing, larger file).
+.TP
+.BI \-restart " int
+Emit JPEG restart markers (\fBRST0\fP..\fBRST7\fP) every \fIint\fP rows of
+MCUs, along with the corresponding \fBDRI\fP header. Restart markers reset
+the entropy coder at regular intervals, which limits the damage a corrupted
+bitstream can do and allows a decoder to process intervals independently, at
+the cost of a slightly larger file. The default is \fB0\fP, which disables
+them. Note the interval is expressed in MCU \fIrows\fP, as with cjpeg's
+option of the same name.
 
 .SH BUGS
 Please report any bugs to the SJPEG discussion list:
--- a/src/dichotomy.cc
+++ b/src/dichotomy.cc
@@ -93,6 +93,11 @@ void Encoder::StoreRunLevels(DCTCoeffs* coeffs) {
   ResetDCs();
   nb_run_levels_ = 0;
   int16_t* in = in_blocks_;
+  // Restart markers reset the decoder's DC predictors, so the dc_code_ deltas
+  // stored here must be relative to the very same boundaries that
+  // FinalPassScan() will later emit the markers at.
+  const int mcus_per_interval =
+      (restart_interval_rows_ > 0) ? restart_interval_rows_ * mb_w_ : 0;
   for (int n = 0; n < mb_w_ * mb_h_; ++n) {
     if (!CheckBuffers()) return;
     for (int c = 0; c < nb_comps_; ++c) {
@@ -106,6 +111,9 @@ void Encoder::StoreRunLevels(DCTCoeffs* coeffs) {
         ++coeffs;
         in += 64;
       }
+    }
+    if (mcus_per_interval > 0 && (n + 1) % mcus_per_interval == 0) {
+      ResetDCs();
     }
   }
 }
@@ -196,6 +204,7 @@ void Encoder::LoopScan() {
     if (ok_) {
       WriteDQT();
       WriteSOF();
+      WriteDRI();
       WriteDHT();
       WriteSOS();
       FinalPassScan(nb_mbs, base_coeffs);
@@ -230,6 +239,7 @@ size_t Encoder::HeaderSize() const {
   size += 8 + 3 * nb_comps_ + 2;  // SOF
   size += 6 + 2 * nb_comps_ + 2;  // SOS
   size += 2;                      // EOI
+  if (restart_interval_rows_ > 0) size += 6;  // DRI
   // DHT:
   for (int c = 0; c < (nb_comps_ == 1 ? 1 : 2); ++c) {   // luma, chroma
     for (int type = 0; type <= 1; ++type) {               // dc, ac
@@ -293,6 +303,13 @@ float Encoder::ComputeSize(const DCTCoeffs* coeffs) {
     BitCounter bc;
     BlocksSize(mb_w_ * mb_h_ * mcu_blocks_, coeffs, all_run_levels_, &bc);
     size += bc.Size();
+  }
+  if (restart_interval_rows_ > 0) {
+    // Each restart costs a 2-byte RSTn marker, plus the '1'-bits padding the
+    // preceding entropy data to a byte boundary (0..7 bits, ~4 on average).
+    const int nb_restarts =
+        (mb_h_ + restart_interval_rows_ - 1) / restart_interval_rows_ - 1;
+    size += nb_restarts * (2 * 8 + 4);
   }
   return size / 8.f;
 }
--- a/src/enc.cc
+++ b/src/enc.cc
@@ -328,6 +328,21 @@ void Encoder::CollectCoeffs() {
 ////////////////////////////////////////////////////////////////////////////////
 // 1-pass Scan
 
+void Encoder::EmitRestartMarker(int interval_idx) {
+  // No Reserve() is needed here: every caller runs inside a scan loop that has
+  // just called CheckBuffers(), which reserves more room than the largest
+  // possible MCU can consume, leaving ample slack for these 2 bytes.
+  bw_.Flush();
+  const uint8_t rst_marker[2] = {
+      0xff, static_cast<uint8_t>(0xd0 + (interval_idx & 7))};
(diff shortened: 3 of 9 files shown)


man/sjpeg.1
```diff
@@ -176,6 +176,15 @@ optimized for rate-distortion).
...
 .B \-no_optim
 Disable Huffman code optimization (faster processing, larger file).
+.TP
+.BI \-restart " int
+Emit JPEG restart markers (\fBRST0\fP..\fBRST7\fP) every \fIint\fP rows of
+MCUs, along with the corresponding \fBDRI\fP header. Restart markers reset
+the entropy coder at regular intervals, which limits the damage a corrupted
+bitstream can do and allows a decoder to process intervals independently, at
+the cost of a slightly larger file. The default is \fB0\fP, which disables
+them. Note the interval is expressed in MCU \fIrows\fP, as with cjpeg's
```
skal65535: not sure we want to mention cjpeg here.

src/dichotomy.cc
```diff
@@ -107,6 +112,9 @@ void Encoder::StoreRunLevels(DCTCoeffs* coeffs) {
         in += 64;
       }
     }
+    if (mcus_per_interval > 0 && (n + 1) % mcus_per_interval == 0) {
```
skal65535: should this test be at the start of the loop instead (with 'n' instead of 'n+1') for clarity?

src/enc.cc
```diff
@@ -358,10 +379,19 @@ void Encoder::FinalPassScan(size_t nb_mbs, const DCTCoeffs* coeffs) {
...
+  const size_t blocks_per_interval =
+      (restart_interval_rows_ > 0)
+          ? static_cast<size_t>(restart_interval_rows_) * mb_w_ * mcu_blocks_
+          : 0;
+  int interval_idx = 0;
   for (size_t n = 0; n < nb_mbs; ++n) {
     if (!CheckBuffers()) return;
     CodeBlock(&coeffs[n], run_levels);
     run_levels += coeffs[n].nb_coeffs_;
+    if (restart_interval_rows_ > 0 && (n + 1) % blocks_per_interval == 0 &&
```
skal65535: equivalent, but safer: 'if (blocks_per_interval > 0 && (n + 1) % blocks_per_interval == 0 && ...)'

I wonder if this should be moved at the beginning of the loop, too (with n>0 condition and 'n+1' replaced by 'n')

src/enc.cc
```diff
@@ -408,6 +438,10 @@ void Encoder::SinglePassScanOptimized() {
         }
       }
     }
+    if (restart_interval_rows_ > 0 &&
+        (mb_y + 1) % restart_interval_rows_ == 0) {
+      ResetDCs();
```
skal65535: ditto: beginning of loop?

src/enc.cc
```diff
@@ -451,6 +485,15 @@ bool Encoder::Encode() {
...
   mb_x_max_ = W_ / block_w_;
   mb_y_max_ = H_ / block_h_;
+  // DRI stores the interval as a 16-bit count of MCUs, so a row-based interval
+  // can't exceed 0xffff MCUs. Clamp to the coarsest whole number of MCU rows
+  // that fits: restarts become more frequent than asked for, never less, and
+  // stay row-aligned as the scan loops and the DC resets require. Dimensions
+  // are capped at kMaxDimension, so mb_w_ <= 8192 and this is always >= 7.
+  if (restart_interval_rows_ > 0) {
+    const int max_rows = 0xffff / mb_w_;
+    if (restart_interval_rows_ > max_rows) restart_interval_rows_ = max_rows;
```
skal65535: restart_interval_rows_ = std::min(restart_interval_rows_, max_rows) ?

src/sjpeg.h
```diff
@@ -216,6 +216,12 @@ struct EncoderParam {
   bool adaptive_bias;           // if true, use perceptual bias adaptation
   bool use_trellis;             // if true, use trellis-based optimization
 
+  // Emit restart markers (RST0-RST7) every 'restart_interval_rows' MCU rows.
+  // Any value <= 0 (the default) disables them. Note the unit is MCU *rows*:
+  // libjpeg's cinfo.restart_interval counts MCUs instead (its row-based
```
skal65535: here too, i'd keep the details about libjpeg too the minimum, if at all

tests/unit_test.cc
```diff
@@ -826,6 +826,222 @@ SJPEG_TEST(Progressive) {
...
+  // Number of MCUs per row, which is what DRI counts: 16x16 MCUs in 4:2:0,
+  // 8x8 otherwise.
+  auto MCUsPerRow = [](int w, SjpegYUVMode yuv_mode) {
+    const int block_w = (yuv_mode == SJPEG_YUV_420) ? 16 : 8;
+    return (w + block_w - 1) / block_w;
+  };
+
+  // Test 4: Decodability across resolutions, intervals, YUV modes, and Huffman
+  // modes. DRI must hold the exact MCU count matching the requested rows.
+  const int dimensions[][2] = {{16, 16}, {64, 64}, {127, 93}, {320, 240}};
```
skal65535, line 903: can you add the maximum stress test to exercise the 16bit counters: input image height = 65536, and reset-interval-rows=1

tests/unit_test.cc
```diff
@@ -826,6 +826,222 @@ SJPEG_TEST(Progressive) {
...
+
+  // Test 6: DRI's interval field is 16 bits, so a row-based interval that
+  // would exceed 0xffff MCUs is clamped down to whole rows. It must never
+  // wrap (which silently desynchronizes the decoder) nor reach zero (which
+  // means 'no restarts' while markers are still being emitted). The image is
+  // sized so that the clamped interval is still shorter than the picture,
+  // hence markers really are emitted and the clamp has to reach the scan
+  // loops, not merely the header.
+  {
+    const int w = 2048, h = 2048;  // 4:0:0 -> 256 MCUs/row, 256 MCU rows
```
skal65535, line 973: i'd even go extreme heree: h=65536, w=something-reasonable

man/sjpeg.1
hasinoff: Done, reference removed.

src/sjpeg.h
hasinoff: Done, trimmed to two lines.

src/enc.cc
hasinoff: Both done. Guarding the actual divisor is clearer, and hoisting makes the n + 1 < nb_mbs term unreachable, so it is gone.

tests/unit_test.cc
hasinoff: Added, at 65535 - kMaxDimension is 0xffff, so 65536 is rejected. Tallest picture at one row per interval gives 8192 intervals / 8191 markers.Added, at 65535 - kMaxDimension is 0xffff, so 65536 is rejected. Tallest picture at one row per interval gives 8192 intervals / 8191 markers.

tests/unit_test.cc
hasinoff: Covered by the same test's second case: widest picture, 8192 MCUs/row, so 7 rows is the largest interval that fits DRI and 8 rows clamps back to exactly 7.

src/enc.cc
hasinoff: Done (needed \<algorithm\>).

src/enc.cc
hasinoff: Done, and applied to all four sites (SinglePassScan, FinalPassScan, SinglePassScanOptimized, StoreRunLevels) so they read alike. Two of them were also doing a pointless final ResetDCs().

src/dichotomy.cc
hasinoff: Done. StoreRunLevels() was also resetting the DC predictors once more after the very last MCU; hoisting drops that too.

tests/unit_test.cc
skal65535: ack

skal65535 (approved):
Verified OK. small fix-up in commit b22de6e:

* restart markers are no-op for progressive encoding, so i added a warning in sjpeg CLI (and sjpeg.h) about it
* bumped the version to 0.1.3 since it's a new feature / CLI flag
* fixed the date in the man page

PR #193, opened 2026-09-15
quantize: introduce -rdo option for fast rate-distortion optimization

skal65535:
rdo does 2 (cheap) things:
  * if the last level is '1', try to set it to 0 if the rd-cost is
    favorable. This is basically performing the first step of the trellis
    optimisation, hard-coded.
  * within the coeffs, if a level is 1, evaluate setting it to 0 and
    merging the run before and after into a single long-run. This is
    trellis-optim, hardcoded for level==1 specifically.

These two special cases of trellis-optim capture ~80% of the efficiency
of the trellis-optim for a mere fraction of its cost.

Gains:
  * saves up to ~6% at q=50, ~2.5% at q=75
  * ~10% encode time over baseline
  * runs 1.3x to 1.6x faster than trellis
  * disto drop: ~0.16 dB Y-PSNR at q=50, <=0.04 dB at q>=85

Bumped version to 0.1.4, since there's a new 'use_rdo' field in the
API (and -rdo option to sjpeg/vjpeg)


src/sjpeg.h
```diff
@@ -24,7 +24,7 @@
 #include <string>
 #include <vector>
 
-#define SJPEG_VERSION 0x000103   // 0.1.3
+#define SJPEG_VERSION 0x000104   // 0.1.4
```
jzern, line 27: Later if we want to say this follows semver.org, anything `0.x` can be a breaking change, so `0.2.0` would conform.

man/sjpeg.1
jzern: This can be updated.

src/quantize.cc
```diff
@@ -321,6 +321,133 @@ static int QuantizeBlock(const int16_t in[64], int idx,
...
+static void MergeRuns(const int16_t in[64], const Quantizer* const Q,
+                      DCTCoeffs* const out, RunLevel* const rl) {
+  int nb = out->nb_coeffs_;
+  const uint32_t* const codes = Q->codes_;
+  if (nb <= 1 || codes == nullptr) return;
+
+  const uint32_t zrl_len = codes[0xf0] & 0xff;
+  RunCost after = {~0u, rl[nb - 1].run_, out->last_};
+  for (int k = nb - 2; k >= 0; --k) {
+    RunCost before = {~0u, rl[k].run_, after.pos - after.run - 1};
```
jzern, line 384: C++20 will allow designated intializers for clarity. C++17 is still preferred here: https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md.

tests/unit_test.cc
```diff
@@ -1130,6 +1130,112 @@ SJPEG_TEST(RestartMarkers) {
   }
 }
 
+SJPEG_TEST(RateDistortionOptimization) {
+  const int kWidth = 64, kHeight = 64;
```
jzern: Can be `constexpr`

tests/unit_test.cc
```diff
@@ -1130,6 +1130,112 @@ SJPEG_TEST(RestartMarkers) {
...
+  const std::vector<uint8_t> rgb = MakeRGB(kWidth, kHeight);
+
+  // 1. Basic encoding across quality factors with use_rdo.
+  const float kQualities[] = {10.0f, 50.0f, 75.0f, 90.0f, 95.0f, 100.0f};
```
jzern: This can be `static constexpr`

tests/unit_test.cc
```diff
@@ -1130,6 +1130,112 @@ SJPEG_TEST(RestartMarkers) {
...
+  for (size_t q = 0; q < ARRAY_SIZE(kQualities); ++q) {
```
jzern: `for (size_t q  : kQualities) {` should work.

tests/unit_test.cc
```diff
@@ -1130,6 +1130,112 @@ SJPEG_TEST(RestartMarkers) {
...
+    sjpeg::EncoderParam param(kQualities[q]);
+    param.use_rdo = true;
+    std::string out;
+    SJPEG_CHECK(EncodeRGB(rgb, kWidth, kHeight, param, &out));
+    SJPEG_CHECK(HasSize(out, kWidth, kHeight));
+  }
+
+  // 2. All YUV modes.
+  const SjpegYUVMode kModes[] = {SJPEG_YUV_AUTO, SJPEG_YUV_420, SJPEG_YUV_SHARP,
```
jzern: `static constexpr` + range based for loop

jzern (approved):
Just some cosmetics.

src/sjpeg.h
skal65535: it's an extra feature, not really breaking. We can go 0.2 when the incoming multi-threading feature is in.

man/sjpeg.1
skal-aa: done

tests/unit_test.cc
skal-aa: done

tests/unit_test.cc
skal-aa: done

tests/unit_test.cc
skal-aa: done

tests/unit_test.cc
skal-aa: done

src/quantize.cc
skal65535: ack

PR #194, opened 2026-09-15
quantize: optimize QuantizeBlockNEON with 2x unrolling and streamlined zigzag loop

skal65535:
### Summary
Optimizes `QuantizeBlockNEON` in `src/quantize.cc` for ARM/NEON architectures and shares streamlined zigzag emission with SSE2.

### Key Changes
• **2x Loop Unrolling**: Process 16 coefficients per iteration in the NEON quantization loop, combining dual 8-lane bit weights (`weights0` and `weights1`) into a single 16-bit lane reduction (`vaddvq_u16(vorrq_u16(nz0, nz1))`), halving vector-to-scalar reduction stalls from 8 to 4 per block.
• **Factorized 8-Lane Helper**: Factorize vector math with an inline lambda to eliminate code duplication across the unrolled loop.
• **Unified `EmitRunLevels`**: Extracted shared helper for both SSE2 and NEON to unify zigzag bitmask remapping, eliminate redundant `dc` / `out->idx_` writes, and share fast exit logic across architectures.
• **Empty AC Fast Exit**: Blocks with no non-zero AC coefficients (`nzn & ~1ull == 0`) return immediately without entering the serialization loops.
• **Streamlined Zigzag Loop**: Pre-compute `levels[i]` during the zigzag bitmask generation loop, eliminating `kZigzag[i]` lookups, `CalcLog2`, bitmasking, and stack array reads from the final serialization loop.

### Performance Gains (on Apple Silicon ARM64, 45,000 blocks)
• $q=50$: 0.64 ms $\to$ 0.59 ms (+7.4%, 1.08x faster)
• $q=75$: 0.78 ms $\to$ 0.74 ms (+5.2%, 1.05x faster)
• $q=85$: 1.03 ms $\to$ 0.95 ms (+7.3%, 1.08x faster)
• $q=95$: 2.00 ms $\to$ 1.86 ms (+6.7%, 1.07x faster)

### Bit-Exactness
100% bit-exact match across all test fixtures, qualities, and the full CLIC image corpus. All unit tests pass.


src/quantize.cc
```diff
@@ -224,66 +243,50 @@ static int QuantizeBlockNEON(const int16_t in[64], int idx,
...
-    const uint16x8_t nz = vandq_u16(vtstq_u16(F, F), weights);
+    vst1q_u16(tmp + offset, F);
+    vst1q_u16(masked + offset, G);
+    return vandq_u16(vtstq_u16(F, F), w);
+  };
+
+  for (int i = 0; i < 64; i += 16) {
+    const uint16x8_t nz =
+        vorrq_u16(quant8(i, weights0), quant8(i + 8, weights1));
 #if defined(__aarch64__)
```
jzern, line 278: This pattern comes up a few times. Could use a horizontal add helper later.

jzern (approved):

src/quantize.cc
skal65535: i'm considering adding it to PR #201 set of macros, actually...

PR #195, opened 2026-09-15
jpeg_tools: NEON accelerate ScoreRow

skal65535:
Vectorizes neutral-chroma (`gray_num`) accumulation in `ScoreRow` (`src/jpeg_tools.cc`) using NEON 8 pixels at a time, and reads row samples directly via pointer offsets.

- **Performance**: +23% faster on `ScoreRow` (Apple Silicon ARM64), ~2% faster on end-to-end `SjpegRiskiness`.
- **Bit-exact**: 100% bit-exact across all test fixtures and the CLIC image corpus.


jzern (approved):

skal65535:
turns out, the original speed-up code wasn't that fast. It was even slower for some input.

=> rewrote the code with a table, like the AVX2 implementation did. And now we're really getting a stable ~4% speedup end-to-end on my M1.

src/jpeg_tools.cc
```diff
@@ -281,6 +283,109 @@ extern void RiskinessScoreRowAVX2(const uint16_t* row1, const uint16_t* row2,
                                   int size, int noise_level,
                                   int64_t* score_sum, int64_t* score_num,
                                   int64_t* gray_num);
+#elif defined(SJPEG_USE_NEON)
+// Row-pointer table into kSharpnessScore: NEON has no gather instruction,
+// so each 2D lookup becomes a plain pointer dereference instead. Mirrors
+// the AVX2 variant in riskiness_avx2.cc.
+static std::vector<const uint8_t*> kRowTableNEON;
```
jzern: There is guidance against non-trivially destructible globals. Seems the lookup might be done inline rather than through this table.

https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

jzern (approved):

src/jpeg_tools.cc
skal65535: good call! i rewrote the function using 8-way index calculation instead of table lookup => 15% faster instead of 4%. I'll see about doing the same for AVX2 later, maybe it's worth it there too.

PTAL

src/jpeg_tools.cc
```diff
@@ -281,6 +282,102 @@ extern void RiskinessScoreRowAVX2(const uint16_t* row1, const uint16_t* row2,
...
+  *v_gray_cnt = vsubq_s16(*v_gray_cnt, vreinterpretq_s16_u16(is_gray));
+
+  // row1/row2 already hold packed 3-channel indices in [0, kRGB3), so
+  // idx_a + kRGB3 * idx_b can reach ~kRGB3^2 (117648 for kRGBSize=7):
+  // needs a 32-bit widening multiply-accumulate, not a 16-bit one.
+  const uint16x4_t r1_lo = vget_low_u16(r1), r1_hi = vget_high_u16(r1);
+  const uint16x4_t r1s_lo = vget_low_u16(r1s), r1s_hi = vget_high_u16(r1s);
+  const uint16x4_t r2_lo = vget_low_u16(r2), r2_hi = vget_high_u16(r2);
+  const uint32x4_t idx0_lo = vmovl_u16(r1_lo), idx0_hi = vmovl_u16(r1_hi);
+  const uint32x4_t idx1_lo = vmovl_u16(r1s_lo), idx1_hi = vmovl_u16(r1s_hi);
```
jzern: I can't think of a way to fold this widening in. You'd still have a widening multiply and add to do instead.

src/jpeg_tools.cc
```diff
@@ -281,6 +282,102 @@ extern void RiskinessScoreRowAVX2(const uint16_t* row1, const uint16_t* row2,
...
+  STORE_16(vmlal_n_u16(idx0_lo, r1s_lo, kRGB3), tmp_ab + 0);
+  STORE_16(vmlal_n_u16(idx0_hi, r1s_hi, kRGB3), tmp_ab + 4);
+  STORE_16(vmlal_n_u16(idx0_lo, r2_lo, kRGB3), tmp_ac + 0);
+  STORE_16(vmlal_n_u16(idx0_hi, r2_hi, kRGB3), tmp_ac + 4);
+  STORE_16(vmlal_n_u16(idx1_lo, r2_lo, kRGB3), tmp_bc + 0);
+  STORE_16(vmlal_n_u16(idx1_hi, r2_hi, kRGB3), tmp_bc + 4);
+
+  auto score_at = [&](int k) -> uint16_t {
+    return (uint16_t)(kSharpnessScore[tmp_ab[k]] + kSharpnessScore[tmp_ac[k]] +
+                       kSharpnessScore[tmp_bc[k]]);
```
jzern: Indent is off.

jzern (approved):

src/jpeg_tools.cc
skal65535: fixed

src/jpeg_tools.cc
skal65535: agreed. We're mixing widen-mults and widen-adds, there might be better combo

PR #196, opened 2026-09-16
enc: fix ctor initializer order for use_rdo_

skal65535:
Fixes:
```
src/enc.cc:84:5: warning: field 'memory_hook_' will be initialized after field 'use_rdo_' [-Wreorder-ctor]
```

`use_rdo_` is declared before `ok_`/`bw_`/etc. in `Encoder` (sjpegi.h), but was initialized last in `Encoder`'s member-init list. Moved it to match declaration order.


skal65535:
missed left-over, sorry

jzern:
lgtm

PR #197, opened 2026-09-18
Optimize Sharp YUV with AVX2 SIMD

hasinoff:
Accelerates Sharp YUV preprocessing and iterative refinement routines in src/yuv_convert_avx2.cc using 256-bit AVX2 SIMD. This vectorizes bicubic row filtering (SharpFilterRow), RGB-to-luma conversion (StoreGray), input row unpacking (ImportOneRow), and the iterative solver loops (SharpUpdateY, SharpUpdateRGB), while eliminating gather instructions in linear-gamma conversion (UpdateW, UpdateChroma) via scalar L1D loads and in-register permutation.

### Isolated SIMD Kernel Microbenchmarks (Clang -O2 -mavx2, x86_64)

- ImportOneRow:      3.50 us ->   0.50 us ( 1.64 -> 11.57 GB/s, +603.9% / 7.04x)
- StoreGray:         2.98 us ->   0.29 us ( 3.86 -> 39.46 GB/s, +921.6% / 10.22x)
- SharpFilterRow:    7.49 us ->   0.35 us ( 2.05 -> 44.08 GB/s, +2049.5% / 21.49x)
- SharpUpdateY:      0.49 us ->   0.11 us (15.77 -> 68.63 GB/s, +335.2% / 4.35x)
- SharpUpdateRGB:    0.04 us ->   0.05 us (88.41 -> 82.29 GB/s,  parity)
- UpdateW:           5.63 us ->   3.69 us ( 2.05 ->  3.12 GB/s,  +52.5% / 1.52x)
- UpdateChroma:      8.83 us ->   6.47 us ( 2.61 ->  3.56 GB/s,  +36.5% / 1.36x)

Full Sharp YUV Preprocessing Stage (ConvertSharpYUV, 1080p):
- 53.44 ms -> 26.39 ms ( 38.8 ->  78.6 MP/s, +102.5% speedup / 2.02x)

Full Sharp YUV Encode (1080p e2e, Method 4, Q85):
- 76.95 ms -> 34.01 ms ( 26.9 ->  61.0 MP/s, +126.2% speedup / 2.26x)

### End-to-End Encoding Throughput (Open Images Corpus, Clang -O2 -mavx2)

Evaluated on 200 full-resolution Open Images (1273.54 Megapixels):

```
Mode / Pipeline                      Base MP/s   AVX2 MP/s   Speedup    Notes
----------------------------------------------------------------------------------------------------------------------
Sharp YUV (Method 4, Q85)              45.39       62.89     +38.55%    Accelerates Sharp YUV preprocessing and solver
AUTO Mode (Method 4, SjpegRiskiness)   66.17       84.29     +27.38%    Accelerates images routed to Sharp YUV
Method 4 Default (Adaptive Q85)       190.86      190.86       0.00%    Untouched (0.0% code diff in non-SharpYUV modes)
```

### Bit-Exactness & Correctness Verification

- 100.0% bit-exact parity verified across all 200 Open Images (1273.54 MP) with 0 byte diffs.
- 100.0% bit-exact parity verified across all 2,663 full-resolution Open Images (~16.4 GP) across Sharp YUV, AUTO, and Default modes with 0 byte diffs.
- All 20 unit test suites pass (3,868 checks, 0 failures).


src/yuv_convert_avx2.cc
```diff
@@ -12,38 +12,316 @@
...
+
+    const __m256i y_packed = _mm256_packs_epi32(y_lo, y_hi);
+    _mm256_storeu_si256(reinterpret_cast<__m256i*>(y + i), y_packed);
+  }
+  for (; i < w; ++i) {
+    y[i] = RGBToGray(rgb[0 * w + i], rgb[1 * w + i], rgb[2 * w + i]);
+  }
+}
+
+static inline void RGB24ToPlanar8(const uint8_t* const rgb, __m128i* const r,
```
skal65535, line 226: would be interesting to compare to the versions we have in libwebp (RGBPackedToPlanar_SSE2, etc.):
https://github.com/webmproject/libwebp/blob/main/src/dsp/yuv_sse2.c#L491

src/yuv_convert_avx2.cc
```diff
@@ -12,38 +12,316 @@
...
+      46871 - 32768, 13933, 46871 - 32768, 13933, 46871 - 32768, 13933,
+      46871 - 32768, 13933, 46871 - 32768, 13933, 46871 - 32768, 13933,
+      46871 - 32768, 13933, 46871 - 32768, 13933);
+  const __m256i mult_b = _mm256_set1_epi32(4732);
+  const __m256i rounder = _mm256_set1_epi32(1 << 15);
+  const __m256i zero = _mm256_setzero_si256();
+
+  for (; i + 16 <= w; i += 16) {
+    const __m256i r =
+        _mm256_loadu_si256(reinterpret_cast<const __m256i*>(rgb + 0 * w + i));
```
skal65535, line 187: fyi: this could be a LOAD_32() macro (we have one in histogram_avx2.cc).

=> i'll do a pass collecting all the SIMD macro into a .h in another PR

src/yuv_convert_avx2.cc
```diff
@@ -12,38 +12,316 @@
...
+    const __m256i g = _mm256_set_m128i(g1, g0);
+    const __m256i b = _mm256_set_m128i(b1, b0);
+    const __m256i r_up = _mm256_or_si256(_mm256_slli_epi16(r, SFIX), shalf);
+    const __m256i g_up = _mm256_or_si256(_mm256_slli_epi16(g, SFIX), shalf);
+    const __m256i b_up = _mm256_or_si256(_mm256_slli_epi16(b, SFIX), shalf);
+    _mm256_storeu_si256(reinterpret_cast<__m256i*>(dst + 0 * w + i), r_up);
+    _mm256_storeu_si256(reinterpret_cast<__m256i*>(dst + 1 * w + i), g_up);
+    _mm256_storeu_si256(reinterpret_cast<__m256i*>(dst + 2 * w + i), b_up);
+  }
+  for (; i < pic_width; ++i) {
```
skal65535, line 291: maybe we should call the ImportOneRow_C() version for the tail?
That's less ad-hoc code, at the expense at exposing the ImportOneRow_C symbol

src/yuv_convert_avx2.cc
```diff
@@ -12,38 +12,316 @@
...
 // C-version for left-over tails.
 extern uint32_t GammaToLinear(int v);
 extern uint32_t LinearToGamma(uint32_t value);
-extern uint32_t RGBToGray(uint32_t r, uint32_t g, uint32_t b);
 extern uint32_t ScaleDown(int a, int b, int c, int d);
 
 //------------------------------------------------------------------------------
 
 static inline __m256i GammaToLinear8(__m256i idx) {
-  return _mm256_i32gather_epi32(
```
skal65535, line 45: the 'gather' version wasn't faster?

skal65535 (commented):
I ran some bench on Haswell with GCC 15 -O3 and Clang 21 -O2:

* the gain on SharpFilterRow/StoreGray/ImportOneRow/SharpUpdateRGB is confirmed
* but removal of 'gather' in GammaToLinear8 / LinearToGamma8 is a big speed regression

The best combo seems to be: leave UpdateW and UpdateChroma with 'gather', and keep the other new functions.

src/yuv_convert_avx2.cc
skal65535: here's the results, and libwebp's version seems much slower:
image

src/yuv_convert_avx2.cc
hasinoff: Done - to handle dst buffer strides cleanly, I added a `start_x` offset param.

hasinoff:
> I ran some bench on Haswell with GCC 15 -O3 and Clang 21 -O2:
> 
> * the gain on SharpFilterRow/StoreGray/ImportOneRow/SharpUpdateRGB is confirmed
> * but removal of 'gather' in GammaToLinear8 / LinearToGamma8 is a big speed regression
> 
> The best combo seems to be: leave UpdateW and UpdateChroma with 'gather', and keep the other new functions.

Thanks for testing on Haswell - my testing on newer cores showed ~1.5x–2x speedup vs. a regression with vpgatherdd. Looking deeper, the regression for the gather-free version on Haswell was very likely caused by STLF stalls - on Haswell, slicing a 256-bit vector store with narrow scalar loads causes a 10-15 cycle stall per element.

In the updated commit we bypass the stack completely, reading directly from the the input image buffer (which is already hot in L1 cache). Could you re-test the updated commit on your Haswell setup to see if the regression is gone? I don't know if it's worth the complexity of maintaining 2 implementations and conditionally compiling or doing cpuid queries for older microarchs.

skal65535:
Re-ran on Haswell (GCC 15 -O3 and Clang 21 -O2) with ca1af8b: the regression is gone.

  - UpdateW/UpdateChroma are now within a few % of the vpgatherdd version, vs +40–60% slower with the previous
    store+reload version:
    - UpdateW: parity on GCC, 11% faster on Clang
    - UpdateChroma: 2–4% slower
  - Sharp YUV stage vs main:
    - +15% (GCC)
    - +8.5% (Clang), also ~4% faster than keeping gather
  - End-to-end on 200 Open Images (1207 MP, m4 q85):
    - SharpYUV: +12.1% (GCC) / +4.2% (Clang)
    - AUTO: +2–4%
  - Bit-exact vs main on all 200 images in SharpYUV and AUTO modes, both compilers; tests pass.
 
Agreed there's no need for two implementations or cpuid gating: the gather-free v2 is fine on Haswell too, and we're talking sub-1% difference here. So let's safe the code maintenance complexity.

skal65535 (approved):

PR #198, opened 2026-09-19
enc: add SafeMultiply and SafeAdd arithmetic helpers

skal65535:
Add overflow-checked `SafeMultiply<T>` and `SafeAdd<T>` templates in `src/sjpegi.h` using compiler builtins (`__builtin_mul_overflow`, `__builtin_add_overflow`) when available, with portable standard C++ fallback.

Calls like `SafeMultiply<size_t>(w, h, &out)` automatically convert arguments without requiring verbose `static_cast` at call sites.

Applies safe sizing to:
- `Encoder::Alloc`
- `Encoder::AllocateBlocks` and `Encoder::Encode` in `src/enc.cc`
- `EncoderSharp420` buffer sizing in `src/encoders.cc`

Adds unit test coverage in `tests/unit_test.cc`.


src/enc.cc
```diff
@@ -287,8 +287,12 @@ bool Encoder::CheckBuffers() {
 bool Encoder::AllocateBlocks(size_t num_blocks) {
   assert(in_blocks_ == nullptr);
   have_coeffs_ = false;
-  const size_t size = num_blocks * 64 * sizeof(*in_blocks_);
-  in_blocks_base_ = Alloc<uint8_t>(size + ALIGN_CST);
+  size_t size = 0;
+  if (!SafeMultiply<size_t>(num_blocks, 64 * sizeof(*in_blocks_), &size) ||
```
jzern, line 291: Automatic type deduction might work here and some of the other cases since the input is `size_t`, so maybe no need for the explicit `<size_t>`.

src/sjpegi.h
```diff
@@ -90,6 +92,82 @@
 
 namespace sjpeg {
 
+////////////////////////////////////////////////////////////////////////////////
+// Safe integer arithmetic helpers for buffer sizing and allocations.
+
+#if (defined(__GNUC__) && __GNUC__ >= 5) || \
+    (defined(__clang__) && defined(__has_builtin) && \
```
jzern, line 99: This will expand all at once, so might fail with some preprocessors. Defining a `SJPEG_HAS_BUILTIN` or doing a two-tiered check is safer. In this case though the fallback will probably be easier since there's the extra `__GNUC__` checks.

```
#if __has_builtin
#if __has_builtin(...)
```

src/sjpegi.h
```diff
@@ -90,6 +92,82 @@
...
+     __has_builtin(__builtin_mul_overflow) && \
+     __has_builtin(__builtin_add_overflow))
+#define SJPEG_HAS_BUILTIN_OVERFLOW 1
+#endif
+
+// Returns true if a * b does not overflow type T, storing the result in *out.
+// If overflow occurs, *out is set to 0 and false is returned.
+template <typename T>
+inline bool SafeMultiply(T a, T b, T* out) {
```
jzern, line 108: (someday) https://en.cppreference.com/cpp/numeric/ckd_mul

jzern (approved):
Nice.

src/sjpegi.h
skal65535: ah! (note: *result is first argument, not last. Would be too easy...)

src/enc.cc
skal65535: yeah, i switched to SafeMultiply<size_t> to avoid the annoying multiple static_cast<size_t> on arguments

src/sjpegi.h
skal65535: ack

PR #199, opened 2026-09-19
enc: add multithreading support via restart intervals

hasinoff:
## Overview

This PR adds multithreading support to `sjpeg` for baseline JPEG encoding using restart intervals (`DRI` / `RST0`..`RST7`). It parallelizes:
1. **Scan Encoding:** Single-pass and two-pass entropy coding across MCU row slices.
2. **Adaptive Quantization (AQ):** Parallel DCT histogram collection across worker threads.
3. **Multi-Pass Dichotomy Search:** Parallel size and PSNR evaluation across dichotomy iterations (`passes > 1`).

All configurations produce byte-identical output to serial encoding for any given restart interval.

## Commit Breakdown (6 Commits)

1. **`enc: add multithreading support via restart intervals`**
   - Adds `EncoderParam::num_threads` (`-threads N` CLI option, `-1` autodetects hardware concurrency).
   - When `num_threads > 1`, `restart_interval_rows` is clamped to at least 1.
   - Factors shared row/slice kernels (`CodeScanSlice`, `QuantizeScanSlice`, etc.) between serial and parallel paths, guaranteeing byte-exact output.
   - Supports rate-distortion optimization (`-rdo`) and Trellis quantization in parallel slices.
   - Conditional compilation via `SJPEG_ENABLE_MULTITHREADING` (CMake), `SJPEG_HAVE_MULTITHREADING` (Makefile / Android.mk).
2. **`enc: parallelize adaptive quantization histogram pass`**
   - Parallelizes `CollectHistograms` across threads with cache-aligned thread-local histogram accumulation, resolving Amdahl's Law bottleneck for Method 4 (2-Pass + AQ).
3. **`enc: parallelize multi-pass dichotomy search (passes > 1)`**
   - Preserves persistent `ThreadPool` across dichotomy passes for both `TARGET_SIZE` and `TARGET_PSNR`.
   - Caches winning run/levels for zero-re-quantization final replay.
4. **`enc: unify serial/parallel slice dispatch via GetNumSlices() and simplify LoopScan`**
   - Unifies serial (1 slice) and parallel (N slices) dispatch through common paths to minimize divergence.
5. **`enc: fix multi-threaded non-reuse methods (2, 6, 8) and UBSAN negative stride`**
   - Supports non-reuse memory modes and fixes pointer arithmetic overflow for bottom-up negative stride buffers under UBSAN.
6. **`enc: use std::unique_ptr for ThreadPool and use clear() over swap()`**
[...]

PR #200, opened 2026-09-19
enc: parallelize ApplySharpYUVConversion via ThreadPool

skal65535:
## Summary
- Parallelizes the 4:2:0 SharpYUV conversion using the Encoder's thread pool, following on from #199.
- Switches the chroma refinement from a serial "Gauss-Seidel" update (row J depends on row J-1 already updated this pass) to a "Jacobi" formulation (all neighbors read from the previous pass), removing the serial dependency chain so slices can be processed in parallel.
- Handles slice-boundary races via thread-local head buffering, flushed after each pass's barrier, guaranteeing bit-exact output vs single-threaded across thread counts.
- Adds `MultiThreadedSharpYUV` unit test; verified under ASan, TSan, and single-threaded, plus manually re-verified bit-exact (1 vs 8 threads) and libjpeg-decodable across all 32 CLIC test images.

Impact on CLIC dataset (q=85): PSNR delta < 0.01 dB, unquantized RGB->YUV PSNR delta < 0.05 dB, compressed size delta < 0.06%, same iteration count to convergence.


skal65535:
rebased and verified against encoding 32 large images (under TSan). Still ok.
End-to-End speed-up is around 50-60% for 4-8 threads.

